PR #9, opened 2015-12-04
Added string helper class

Verox-:


NouberNou:
This needs some consideration as strings in the RV engine are very interesting. They can persist even after their data wrapper is gone, and they can actually persist forever in some cases with no SQF references holding on to them.

For example, try using this helper with the already completed `sideChat` function and see if it presents any errors. I found that `sideChat` holds on to the internal string reference long after the call is made.

If this helper goes out of scope, the string data will also go out of scope, and the char pointer location will be freed, potentially causing heap corruption.

NouberNou:
We still might be able to use a helper here, just for the sake of consistency though, but it wont ever be able to allocate the string solely on the stack. It will always have to allocate the string on the heap, and it will have to call to free the string in its destructor so that the internal Intercept memory handler can monitor it until its reference counter goes back down to 1, in which case that means the RV engine has released all of its references and we can finally free it from our heap.

NouberNou:
Reimplemented in master.

PR #20, opened 2015-12-05
Fixed error in macro declaration

dbenson24:
_WINDOWS was breaking the build. _WIN32 isn't

### DIFF
--- a/src/host/common/shared.hpp
+++ b/src/host/common/shared.hpp
@@ -22,7 +22,7 @@
 #define EXTENSION_RETURN() return;
 #endif
 
-#ifdef _WINDOWS
+#ifdef _WIN32
 #define sleep(x) Sleep(x)
 #endif
 


NouberNou:
Which compiler are you using @dbenson24?

dbenson24:
@NouberNou I am using Visual Studio 2015

NouberNou:
Odd, I believe all of the rest of us are also on 2015 as well and we have not had this issue.

dbenson24:
@NouberNou That really is odd. The _WIN32 vs _WINDOWS directive change won't impact your compilers, could you merge it just so I don't have to change it on my end every time I pull?

Glowbal:
> WIN32 - always defined for Win32 applications
> _WINDOWS - defines that target OS is windows

dbenson24:
http://en.cppreference.com/w/cpp/thread/sleep_for
Could we just switch to std library sleep_for then?

NouberNou:
Fixed in #103

Issue #29, opened 2015-12-15
Incorrect implementation of some autogenerated wrappers


Verox-:
Some of the autogenerated wrappers for the binary functions have inverted parameters, but not all.
Ideally we could just regenerate these wrappers, it'd be the fastest and easiest method of fixing this problem, the other way is doing it manually. (dibs not me)

Glowbal:
I fixed this once, it took ~10 minutes. I will redo it.

jonpas:
Some fixed in #50

NouberNou:
Is this done? I think it is?

jonpas:
There probably still are a few wrong, but we fix them as we see them.

NouberNou:
Cool, closing then.

jonpas:
Remove WIP label as well. :)

PR #31, opened 2015-12-15
Fix inverted binary function arguments

Glowbal:
This fixes most of the inverted binary function arguments. A manual check afterwards is still required.

#29


Glowbal:
Closing in favour of #32

PR #38, opened 2016-01-25
Added ACE3 build.py to create intercept_core.pbo

esteldunedain:
The magic of OSS

### DIFF
--- a/tools/build.py
+++ b/tools/build.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python3
+
+import os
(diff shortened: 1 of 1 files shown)


Glowbal:
Thanks @esteldunedain :+1:

esteldunedain:
To be honest, I was strugging to make the pbo without this.

jonpas:
@esteldunedain You just beat me, I was planning to do that today, we should plan on Slack. :P

PR #39, opened 2016-01-25
Add ACE3 setup.py to prepare dev environment (P drive only)

jonpas:
Removed CBA includes of course.

This is P drive setup only, no CMAKE solution generation.

### DIFF
--- a/tools/setup.py
+++ b/tools/setup.py
@@ -1,0 +1,103 @@
+#!/usr/bin/env python3
+
+############################
+#  Intercept Setup Script  #
+############################
+
+import os
+import sys
+import shutil
+import platform
+import subprocess
+import winreg
+
+######## GLOBALS #########
+MAINDIR = "z"
+PROJECTDIR = "intercept"
+##########################
+
+def main():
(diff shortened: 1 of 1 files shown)


tools/setup.py
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env python3
+
+#######################
+#  ACE3 Setup Script  #
```
esteldunedain: ACE3 by Intercept?

jonpas: Missed that one!

tools/setup.py
```diff
@@ -0,0 +1,103 @@
...
+
+    print("# Links created successfully.")
+    return 0
+
+
+if __name__ == "__main__":
+    exitcode = main()
+
+    if exitcode > 0:
+        print("\nSomething went wrong during the setup. Make sure you run this script as administrator. If these issues persist, please follow the instructions on the ACE3 wiki to perform the setup manually.")
```
esteldunedain: There's one more here

PR #41, opened 2016-01-25
Add near_objects wrapper and alt syntax for nearest_building

jonpas:
My first intercept C++ PR, go easy. :grin: 
(I am totally not copying @BaerMitUmlaut)

Adds wrappers for:
- `nearestBuilding` alt syntax
- `nearObjects` main syntax (position or object)
- `nearObjects` alt syntax (position or object with array of string and number)


NouberNou:
Looks good, yea, making sure to hit all the potential overloads on wrappers is a big thing. Try to make things optional parameters instead of overloads when possible.

PR #42, opened 2016-01-25
Make intercept not load any extension by default.

esteldunedain:
~~Set a global variable when intercept is available and loaded~~

I reverted the added global variable, as there an `intercept_invoker_ok` variable set in the `missionNamespace`.

### DIFF
--- a/rv/addons/core/boot.sqf
+++ b/rv/addons/core/boot.sqf
@@ -8,8 +8,8 @@ _res = "intercept" callExtension format["init_patch:%1", (productVersion select 
     // diag_log text format["Found: %1", _res];
     // ceil 9.5; // this did our patch... probably. :)
     // _res = "intercept" callExtension "remove_patch:"; // clear the patch
-    
-    
+
+
     "intercept" callExtension "invoker_begin_register:";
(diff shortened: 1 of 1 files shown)


rv/addons/core/boot.sqf
```diff
@@ -48,5 +48,6 @@ _res = "intercept" callExtension format["init_patch:%1", (productVersion select
     str parsingNamespace;
     "intercept" callExtension "invoker_end_register:";
 // };
-"intercept" callExtension "load_extension:z\intercept\build\win32\example_frag\RelWithDebInfo\example_frag.dll";
-diag_log text format["_________________________________________Intercept Res: %1", _res];
\ No newline at end of file
+//"intercept" callExtension "load_extension:z\intercept\build\win32\example_frag\RelWithDebInfo\example_frag.dll";
+diag_log text format["_________________________________________Intercept Res: %1", _res];
```
jonpas: This is also unnecessary, `_res` is already logged above and at this point it returns empty string (or nothing).

NouberNou: If you want to comment that out too that is fine.

esteldunedain: Done

PR #45, opened 2016-01-26
Signalling extensions from SQF

esteldunedain:
This PR adds a way of communicating data from SQF to an extension through Intercept.

I'm not sure this is the correct/best way to do it, so please review. 

~~Besides, it's crashing like a test dummy; but apparently most Intercept is doing that today. I'll try to pinpoint the cause.~~


Glowbal:
What would be a use case in where this would be benefitial to have?

esteldunedain:
The use case I was thinking about is to better integrate extensions with existing sqf frameworks: e.g. enabling/disabling features based on the ACE_settings framework, passing custom events, etc.

Alternatively, those things could be done asynchronously through GVARS, but that requires polling the variable all frames from the extension to notice changes.

Glowbal:
Thanks @esteldunedain. I think it's a good things to support this in intercept. It ideally wouldn't be necessary but I can see it offer some flexibility.

@NouberNou will probably take a look at this soon and determine if this is something we would like in intercept.

src/host/invoker/invoker.cpp
```diff
@@ -142,6 +143,25 @@ namespace intercept {
...
+
+        auto modules = extensions::get().modules();
+        if (modules.find(extension_name) == modules.end()) {
+            return false;
+        }
+        auto module = modules[extension_name];
+
+        _signal_params = invoke_raw("getvariable", &_mission_namespace, "NAMESPACE", &game_value("intercept_signal_var"), "STRING");
+        if (module.functions.on_signal) {
+            module.functions.on_signal(_signal_params);
```
Glowbal: May need to check if `_signal_params` is a valid value?

rv/addons/core/config.cpp
```diff
@@ -37,5 +37,13 @@ class CfgFunctions
                 headerType = -1;
             };
         };
+        class api
+        {
+            class signal
+            {
+                file = "z\intercept\rv\addons\core\signal.sqf";
```
overfl0: Shouldn't there be double backslashes?

Itsyuka: @overfl0 You'd think, but the cpp/hpp/h files for Arma 3 doesn't use escapes as we traditionally see them as.

src/client/intercept/intercept.hpp
```diff
@@ -15,6 +15,7 @@ namespace intercept {
         DLLEXPORT void __cdecl mission_end();
         DLLEXPORT void __cdecl mission_stopped();
         DLLEXPORT void __cdecl on_frame();
+        DLLEXPORT void __cdecl on_signal(std::string &signal_name_, game_value& value1_);
```
esteldunedain: I recon we don't want `std::string`s over the ABI, but this string isn't a `rv_string` on the first place and I don't know if we want to a dd yet another string type to Intercept. Maybe add a contructor to make `rv_string`from `std::string`?

esteldunedain: Mmm, would this work?:
```c++
rv_string rv_signal_name(game_value(signal_name));
```

src/host/invoker/invoker.cpp
```diff
@@ -142,6 +143,26 @@ namespace intercept {
...
+        std::string signal_name = args_.as_string(1);
+        LOG(DEBUG) << "Signal " << extension_name << " : " << signal_name << " START";
+
+        auto modules = extensions::get().modules();
+        if (modules.find(extension_name) == modules.end()) {
+            return false;
+        }
+        auto module = modules[extension_name];
+        _invoker_unlock signal_lock(this);
+        _signal_params = invoke_raw("getvariable", &_mission_namespace, "NAMESPACE", &game_value("intercept_signal_var"), "STRING");
```
esteldunedain: I'm still trying to figure out why this line 158 makes Arma CTD. If instead I replace it with:
```c++_signal_params = game_value(std::string("test"));

NouberNou:
Is this working now that the crash bug was fixed?

Also, what would make this even more useful is if you could specify the name of the exported function. I'd suggest using GetProcAddress to return the address of the function the first time and then storing the address in a unordered_map (since looking up in that is going to be faster than GetProcAddress) for subsequent calls.

Also I'd like to investigate how to return variables from this.

NouberNou:
Gunna merge this in and work on adding return variables and user defined function names.

PR #52, opened 2016-01-28
Add not-so-simple nular wrappers

jonpas:


Glowbal:
Functions such as `date` and `fog_params` always return the same type of array. Only it's values can be different. Would it be worth it to have our wrapper methods return a struct instead?

jonpas:
`date` and `fog_params` always return a number, at least that's what the documentation says.

Glowbal:
`date` returns `[year, month, day, hour, minute]`.

You can translate this into a struct:
```c++
struct rv_date {
    float year;
    float month;
    float date;
    float hour;
    float minute;
}
```

jonpas:
Will change to structs, I did not realize this is how this is done... changing title to not-so-simple. :grin:

PR #54, opened 2016-01-28
Added camera function wrappers

BaerMitUmlaut:


src/client/intercept/client/sqf/uncategorized.hpp
```diff
@@ -230,6 +230,32 @@ namespace intercept {
...
+        object cam_create(const std::string &type_, const vector3 &position_);
+        void camera_effect(const object &camera_, const std::string &name_, const std::string &position_);
+        void camera_effect(const object &camera_, const std::string &name_, const std::string &position_, const std::string &rtt_);
+        void cam_prepare_focus(const object &camera_, float distance_, float blur_);
+        void cam_prepare_fov_range(const object &camera_, float min_, float max_);
+        void cam_prepare_pos(const object &camera_, const vector3 &position_);
+        void cam_prepare_rel_pos(const object &camera_, const vector3 &relative_position_);
+        void cam_prepare_target(const object &camera_, const object &target_);
+        void cam_prepare_target(const object &camera_, const vector3 &target_);
+        // Broken command cam_set_dir
```
jonpas: Add the function reference to .cpp so it gets marked in coverage issue.

BaerMitUmlaut: Done, thanks for the hint.

jonpas: Make a note in cpp that it's broken as well, and you can completely remove it from hpp.

PR #56, opened 2016-01-29
Fix Intercept breaking CBA event handlers

esteldunedain:
Also remove unneeded CBA required addon and debug print.

According to @commy2 this solution is still very fragile:
> AllVehicles usually has no EventHandlers class, so this could change inheritance of addons loaded after intercept. And some sub classes of AllVehicles overwrite this EventHandlers class.
> It's basically the same problem [CBA XEH has](https://github.com/CBATeam/CBA_A3/blob/master/addons/xeh/CfgVehicles.hpp).
> And this problem is also the reason why the new EventHandler sub classes cannot replace XEH

An alternative to this is PRed in #57


NouberNou:
Is it possible to mix and match, like using the way CBA handles this but using the BI event handler classes?

esteldunedain:
> Is it possible to mix and match, like using the way CBA handles this but using the BI event handler classes?

@commy2 is the expert on this stuff. Based on what he told me, we can use the same technique CBA uses for its config. But we have to duplicate their maintainance work and even then some addons won't be compatible.

CBA has a fallback for those in the form of an optional PBO that checks all units at regular intervals an injects them with eventhandlers if they are not properly configured, If we choose this route we might want to do the same.

esteldunedain:
Closed in favor of #57.

PR #57, opened 2016-01-29
Make Intercept use XEH

esteldunedain:
This is an alternative solution to that in #56.

It makes Intercept use CBA XEH framework for events. 

## Advantages:
- All the complications derived from addon compatibility are offloaded to CBA.
- Reduced chances of incompatible content existing: if it works with CBA/ACE it works with Intercept
- Requires less maintainance

## Disadvantages:
- CBA is required
- Some extra performance overhead 
- CBA doesn't handle `handleDamage`, probably because of it requiring a return value (Intercept will have to deal with the same problem anyway).
- Some events are not supported by XEH at the moment.

### DIFF
--- a/rv/addons/core/config.cpp
+++ b/rv/addons/core/config.cpp
@@ -540,7 +540,7 @@ class CfgPatches {
             "a3_weapons_f_vests",
             "a3data",
             "map_vr",
-            "extended_eventhandlers"
+            "extended_eventhandlers", "cba_ui", "cba_xeh", "cba_xeh_a3", "cba_jr"
         };
         version = 0.1;
     };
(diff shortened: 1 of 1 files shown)


NouberNou:
I guess we should go with this method for now, and let it be a CBA thing. I don't mind making CBA a requirement either.

NouberNou:
Is this working now that the crash bug was fixed?

Also, what would make this even more useful is if you could specify the name of the exported function. I'd suggest using `GetProcAddress` to return the address of the function the first time and then storing the address in a `unordered_map` (since looking up in that is going to be faster than `GetProcAddress`) for subsequent calls.

Also I'd like to investigate how to return variables from this.

PR #58, opened 2016-01-29
getvariable wrappers

esteldunedain:
These are the wrappers for all the syntaxes of `getvariable`. As far as I tested they work alright.


esteldunedain:
yay!

PR #59, opened 2016-01-30
Adding tree view wrappers

jameslkingsley:
I'll be adding more commits as I go, just here so people can see what's been done already.

### DIFF
--- a/src/client/intercept/client/sqf/uncategorized.cpp
+++ b/src/client/intercept/client/sqf/uncategorized.cpp
@@ -56,7 +56,7 @@ namespace intercept {
         // unary__rad__scalar_nan__ret__scalar_nan
         // unary__random__array__ret__scalar_nan
         // unary__random__scalar_nan__ret__scalar_nan
-        // unary__reverse__array__ret__nothing 
+        // unary__reverse__array__ret__nothing
         // unary__round__scalar_nan__ret__scalar_nan
         // unary__selectrandom__array__ret__any
         // unary__sin__scalar_nan__ret__scalar_nan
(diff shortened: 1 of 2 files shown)


jameslkingsley:
That's all of them. Let me know if I've made any errors; most of the commands were practically identical.

NouberNou:
Most of these look good, except colors can be passed using `sqf::color` like `const rv_color &`. It is a struct with `rgba` fields.

jameslkingsley:
Okay I think that's right now.

esteldunedain:
:+1:

src/client/intercept/client/sqf/uncategorized.cpp
```diff
@@ -6531,5 +6531,234 @@ namespace intercept {
...
+
+            game_value params({
+                path,
+                value_
+            });
+
+            host::functions.invoke_raw_binary(client::__sqf::binary__tvsetvalue__control__array__ret__nothing, ctrl_, params);
+        }
+
+        void tv_sort(const control& ctrl_, const std::vector<int>& path_, const bool& reversed_ = false) {
```
esteldunedain: I think I heard the default parameters should ony be specificied on the .hpp file

src/client/intercept/client/sqf/uncategorized.cpp
```diff
@@ -6531,5 +6531,234 @@ namespace intercept {
...
+
+            game_value params({
+                path,
+                reversed_
+            });
+
+            host::functions.invoke_raw_binary(client::__sqf::binary__tvsort__control__array__ret__nothing, ctrl_, params);
+        }
+
+        void tv_sort_by_value(const control& ctrl_, const std::vector<int>& path_, const bool& reversed_ = false) {
```
esteldunedain: Same situation as above

Glowbal:
:+1: Thanks @jameslkingsley

PR #63, opened 2016-01-30
Travis cmake

jonpas:
It is possible to compile projects that require `windows.h` on Travis with MinGW as it contains the it, source: http://stackoverflow.com/questions/28155308/how-to-cross-compile-with-mingw-on-linux-for-windows

Just testing out step by step.


jonpas:
AppVeyor has been set-up instead, closing in favour of it. Might do Travis builds for Linux when Windows dependency is removed.

PR #65, opened 2016-01-31
Added more ctrl wrappers

jameslkingsley:
I also tested it built without errors/warnings and it did.


jameslkingsley:
I also tested it built without errors/warnings and it did.

jameslkingsley:
Actually it turns out there were warning, I'm obviously blind...

PR #68, opened 2016-02-01
Wrapper autogeneration fixes.

Verox-:


Glowbal:
Nice one @Verox- :+1:

PR #71, opened 2016-02-04
Unary Tree View functions

overfl0:
It's my first PR to Intercept so please review this with care :).

I'm not sure if creating a new __convert_to_integers_vector helper was the right choice or whether I should have stuck with __convert_to_numbers_vector instead.


Glowbal:
CI build is due to failing master. That has been fixed. This is good to go I think :+1:

PR #81, opened 2016-02-09
Add ReSharper team-shared Setting Layer config file

jonpas:
Basically sets the brace style, nothing else. I kept the C-style cast used instead of static_cast as it's only a Suggestion, not marked as an error (and we should probably use static_cast anyways).

Pinging @Verox-, not sure who else uses ReSharper.


Verox-:
Closed for re-pr.

PR #83, opened 2016-02-10
Reshuffled files and made client output as a static library.

Verox-:
Headers are physically split away from the implementation files.
CMakeLists changed across the library to represent the new file structure

The result of this change is to make the client side of the library output as a static library (.lib) for inclusion into clients proper. This way we can distribute our side of the client built, ready for single-file inclusion into the user's mod.


Glowbal:
I like the idea. I will look it over soon.

Glowbal:
Note this PR takes priority over all other wrapper PRs as discussed on Slack.

Verox-:
This isn't confirmed to work yet.

PR #95, opened 2016-02-18
Added path limiting to client module loading.

NouberNou:
Client plugins are now limited to being loaded from specific mod folders, in which they will have a folder called intercept. That folder will contain plugin dlls. Plugins are referred to by their dll name, minus the file extension.

For example:

`"intercept" callExtension "load_extension:example_dll";`

### DIFF
--- a/rv/addons/core/boot.sqf
+++ b/rv/addons/core/boot.sqf
@@ -48,5 +48,24 @@ _res = "intercept" callExtension format["init_patch:%1", (productVersion select 
     str parsingNamespace;
     "intercept" callExtension "invoker_end_register:";
 // };
-//"intercept" callExtension "load_extension:z\intercept\build\win32\example_frag\RelWithDebInfo\example_frag.dll";
+//"intercept" callExtension "load_extension:example_dll";
+
+_intercept_projects = configFile >> "Intercept";
+for "_i" from 0 to (count _intercept_projects)-1 do {
+    _project = _intercept_projects select _i;
+    if(isClass _project) then {
+        for "_x" from 0 to (count _project)-1 do {
+            _module = _project select _x;
+            if(isClass _module) then {
+                _plugin_name = getText(_module >> "pluginName");
+                if(_plugin_name != "") then {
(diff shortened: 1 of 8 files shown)


NouberNou:
My only concern with this is that I ran into a very weird race condition with easylogging++ after using the PBO search tool. I've never seen this in other projects that use this code (ACE VD and ACRE, which has been using it in lots of test deployments for months now). I fixed it by removing a logging call from the constructor of the extensions module, but I am not sure if that actually means anything (it is also where the search happens, but it just seemed odd).

The search code is using a really hacked out NtDll loaded functions search. @jaynus wrote it for ACE VD, and it is mostly black magic Windows API shit to me.

rv/addons/core/boot.sqf
```diff
@@ -48,5 +48,24 @@ _res = "intercept" callExtension format["init_patch:%1", (productVersion select
     str parsingNamespace;
     "intercept" callExtension "invoker_end_register:";
 // };
-//"intercept" callExtension "load_extension:z\intercept\build\win32\example_frag\RelWithDebInfo\example_frag.dll";
+//"intercept" callExtension "load_extension:example_dll";
+
+_intercept_projects = configFile >> "Intercept";
+for "_i" from 0 to (count _intercept_projects)-1 do {
```
jokoho48: why you using here not count/forEach and configClasses? that is faster than for

NouberNou: No real reason, this runs at boot once, so it being super performant is not really that critical.

jaynus:
+1 for win32 internals voodoo

NouberNou:
Yea but you were right, it beats trying to parse every manner of fucking mod path loading.

Though someone will have to write a linux version of this at some point.

jaynus:
That's WAY easier - you just read /proc/<pid>/fd

Yea but you were right, it beats trying to parse every manner of fucking
mod path loading.

Though someone will have to write a linux version of this at some point.

—
Reply to this email directly or view it on GitHub
<https://github.com/intercept/intercept/pull/95#issuecomment-186072367>.

PR #97, opened 2016-03-27
Add various unary wrappers for many new functions since 1.56 and dev

jonpas:


NouberNou:
Looking good, will merge in.

PR #98, opened 2016-03-27
Add unary wrappers with possible different return types

jonpas:


src/client/headers/client/sqf/uncategorized.hpp
```diff
@@ -3184,5 +3207,41 @@ namespace intercept {
...
+            {
+            }
+        };
+
+        struct rv_weapon_items {
+            std::string weapon;
+            std::string muzzle;
+            std::string laser;
+            std::string optics;
+            rv_magazine magazine;
```
jonpas: There might be 2 of these when the grenade launcher is loaded. Not sure how to handle that, I believe @Verox- had an idea?

dedmen: second one std::optional ?

PR #100, opened 2016-05-11
Fix /W4 warnings

jonpas:
Apart from the following `/W4` warning level will now be without warnings.

```
1>waypoint.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>position.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>multiplayer.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>misc.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>marker.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>group.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>eden.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>curator.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>ctrl.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>core.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
1>chat.obj : warning LNK4221: This object file does not define any previously undefined public symbols, so it will not be used by any link operation that consumes this library
```


Glowbal:
Looks fine. I noticed you have some commented code now, which was previously a method but not fully implemented. Perhaps it's better to simply remove those instead?

jonpas:
Done, I am not sure about https://github.com/intercept/intercept/pull/100/files#diff-2351da3a535609edacfd593f44e905e2R94, Visual Studio said it wasn't used, but it might be preparation for something @NouberNou was planning to do.

dedmen:
I don't really like the removal of these functions.
Also pool.hpp will be obsolete after #120 So.. I'd say close this and redo it after #120 is merged. I planned a refactoring run anyway.

dedmen:
I see a couple of misc refactorings. BUt nothing related to the warnings :o

jonpas:
It definitely fixed level 4 warnings when I made this.

PR #103, opened 2016-06-11
Fix compatibility with GCC (MinGW)

marceldev89:
Probably a step in the right direction for Linux compatibility as well. :smiley: 

Includes the same #ifdef _WINDOWS changes as in https://github.com/intercept/intercept/pull/20. `_WIN32` should be defined by all compilers targeting windows.

I'm not sure about the changes for `src/host/invoker/eventhandlers.cpp` in https://github.com/intercept/intercept/commit/bd919bb845788add347ddf0f4e77fe70ba6b582d. But how things were resulted in `error: invalid initialization of non-const reference of type ‘<type>&’ from an rvalue of type ‘<type>’`. MSVC allows it while it shouldn't (http://stackoverflow.com/q/7189420).

Note: all this is untested but compiles with `mingw-w64-i686-gcc 5.4.0` and Visual Studio 2015.


src/client/headers/shared/pool.hpp
```diff
@@ -45,35 +45,35 @@ namespace intercept {
...
                 _buy_entry(Alloc_Count);
         }
 
-        inline T * acquire(size_t alloc_count_) {
+        inline T * acquire(std::size_t alloc_count_) {
             if (_pool_queue.size() == 0 || alloc_count_ > Alloc_Count)
                 _buy_entry(alloc_count_);                
             T * ret = _pool_queue.front();
-            size_t *count = (size_t *)((char *)ret - sizeof(size_t));
+			std::size_t *count = (std::size_t *)((char *)ret - sizeof(std::size_t));
```
jonpas: Tabs instead of spaces.

src/client/headers/shared/pool.hpp
```diff
@@ -45,35 +45,35 @@ namespace intercept {
...
             *count = alloc_count_;
             _pool_queue.pop();
             return ret;
         }
 
         inline void release(T *_ptr) {
-            size_t *count = (size_t *)((char *)_ptr - sizeof(size_t));
-            for (size_t i = 0; i < *count; ++i)
+			std::size_t *count = (std::size_t *)((char *)_ptr - sizeof(std::size_t));
```
jonpas: Tabs instead of spaces.

NouberNou:
Sounds good, will review further soon.

PR #108, opened 2016-11-18
1.66 compat

dedmen:
Compatability update for Arma 1.66
I didn't test in 1.64 but i think it should also work there.
Also somehow lib.sqf is not executing in preInit.. I dunno why. I just called manually. Also that causes a crash if any Eventhandler is called and the invoker is not initialized.
Btw your slackin link is down :x

### DIFF
--- a/src/client/intercept/shared/types.cpp
+++ b/src/client/intercept/shared/types.cpp
@@ -90,12 +90,11 @@ namespace intercept {
             if (single_type != NULL) {
                 return single_type->type_name->string();
             }
-            else if (compound_type != NULL) {
+            else {
                 return
                     compound_type->types->first->type_name->string() + "_" +
                     compound_type->types->second->type_name->string();
             }
-            __debugbreak();//Invalid type
         }
 
         game_data_number::game_data_number() {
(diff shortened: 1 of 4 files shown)


src/client/intercept/shared/types.cpp
```diff
@@ -90,11 +90,12 @@ namespace intercept {
             if (single_type != NULL) {
                 return single_type->type_name->string();
             }
-            else {
+            else if (compound_type != NULL) {
                 return
                     compound_type->types->first->type_name->string() + "_" +
                     compound_type->types->second->type_name->string();
             }
+            __debugbreak();//Invalid type
```
dedmen: Could probably remove this again.. But i thought its better to break instead of crash.

marceldev89: I guess it's not *that* important at this stage but `__debugbreak();` is not portable.

dedmen: MSVC only? Didn't know that.. That is important enough then because people are already using intercept with gcc.

dedmen:
I see you accepted my hidden camelCase classnames :D I'll change them to conform the standarts one day

PR #119, opened 2017-04-17
rv_color order should be RGBA

overfl0:
The color order of rv_color is stored internally as RBGA, which is probably a bug.

This will work fine as long as all the values are passed at once because the `game_value` operator uses the same order as the constructor but if someone wants to access particular color components, they will access the wrong components.

This PR assumes PR #117 has already been merged.


overfl0:
@Glowbal I have fixed the conflicts that have appeared after merging PR #117

PR #120, opened 2017-04-28
Allocator Overhaul

dedmen:
Adds Generic Arma Allocator for RString and Arrays.
Adds Pool Allocator for GameData types

These allocators provide direct access to the Engine Allocator. Meaning we no longer have to wait for our allocated objects to become unused. Because Arma can now deallocate our allocated objects and we can also deallocate Arma objects.

Also includes new r_string implementation that mirrors the real engine type more closely
Also provides correctish Implementation of RefCount class so we don't need to emulate refcounting ourselves anymore.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ rv/intercept.pdb
 !.gitkeep
 vcproj/*
 !vcproj/.gitkeep
+.vs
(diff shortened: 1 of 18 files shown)


dedmen:
The old pool caches are now obsolete. Not sure If I should remove it in this or a seperate PR.

PR #137, opened 2017-05-29
Add setFog wrapper

SilentSpike:
Just familiarising myself with what's going on here, pretty sure I've done this right :smile:

### DIFF
--- a/src/client/headers/client/sqf/world.hpp
+++ b/src/client/headers/client/sqf/world.hpp
@@ -1,4 +1,4 @@
-﻿/*!
+/*!
 @file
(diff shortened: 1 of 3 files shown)


dedmen:
https://github.com/intercept/intercept/pull/134/files#diff-2300cf57dc13d2d11d2cb6b5693fa87aR155
As I said before in ACE Slack. Wrappers are 100% completed right now ^^ Soo.. If you add a wrapper for something you are definetly doing something wrong ^^

But also Intercept is quite chaotic right now because of the huge number of changes going on. You quickly loose overview if you don't read in Slack

PR #140, opened 2017-06-02
Develop->master

dedmen:


NouberNou:
I am down for merging if all other reviewers are?

dedmen:
I want to get the broken C ABI interface out first.
the RSQF stuff uses std::strings there I'm gonna replace these and the other const char*'s with std::string_view. Which is basically a struct with const char* and length so it's fine passing that through.

dedmen:
Neat sideeffect as all get*function_typed now take string_view the lengths of these strings are computed and stored on runtime.
Which should make a huge difference for:
https://github.com/intercept/intercept/blob/master/src/client/headers/client/sqf_assignments.hpp

dedmen:
Also made the RSQF function pointers inaccessible to Users and added wrapper functions instead. This way we can use `[[nodiscard]]` for them. Which we couldn't for the function pointers.

dedmen:
x64/win32 and gcc linux builds all successful. I'll merge this then.

PR #142, opened 2017-06-04
Fix waypoint formation lookup was unable to parse formation values containing underscores

SpyderBlack723:


dedmen:
`NO CHANGE` should not be seperated by _ but the others in that list should be?

SpyderBlack723:
Checked the other lookups but had failed to check the rest of the formations :). The only lookup that should contain an underscore is CIRCLE_L loiter type. Update pushed.

PR #144, opened 2017-06-05
Update requiredVersion to 1.70, Minor PBO core update

jonpas:


rv/addons/core/config.cpp
```diff
@@ -532,15 +801,71 @@ class CfgPatches {
...
             "a3_weapons_f_rifles_sdar",
             "a3_weapons_f_rifles_smg_02",
             "a3_weapons_f_rifles_trg20",
             "a3_weapons_f_rifles_vector",
+            "a3_weapons_f_smgs_pdw2000",
+            "a3_weapons_f_smgs_smg_01",
+            "a3_weapons_f_smgs_smg_02",
             "a3_weapons_f_uniforms",
             "a3_weapons_f_vests",
+            "curatoronly_air_f_beta_heli_attack_01",
```
dedmen: What is this?

jonpas: It is. This was copied directly from ACE3.

PR #145, opened 2017-06-06
Fix pboprefix missing \rv\

jonpas:
- Title
- Add all files in `rv/` folder to .gitignore except `rv/addons/` folder - previously only Win32 build stuff, now catches everything

_Already included in v0.8.0 package._

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,14 +1,12 @@
+rv/*
+!rv/addons
 rv/addons/*.pbo
 rv/addons/*.bat
 rv/*.dll
 rv/carma_dll.*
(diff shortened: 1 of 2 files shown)


dedmen:
I don't know why we even need to have rv there.

jonpas:
We could get rid of if, and change "BUILD2RV" to point to root and place addons folder on root, that's the only reason we have it  from what I see. I don't mind it, and all paths are now using it already.

jonpas:
No one should be relying on internal paths, can always change it down the line.

PR #150, opened 2017-06-10
Added checkVisibility

tbeswick96:
I couldn't find an implementation for [checkVisibility](https://community.bistudio.com/wiki/checkVisibility), so I added it. Just copied implementation from intersects, hope it's ok.

### DIFF
--- a/src/client/headers/client/sqf/intersects.hpp
+++ b/src/client/headers/client/sqf/intersects.hpp
@@ -122,5 +122,8 @@ namespace intercept {
 
         vector3 terrain_intersect_at_asl(const vector3 &pos1_, const vector3 &pos2_);
 
+        float check_visibility(const object& ignore_, sqf_string_const_ref lodname_, const vector3 &begin_pos_, const vector3 &end_pos_);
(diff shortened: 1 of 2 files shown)


dedmen:
@jonpas Did you check that we really don't have it yet? Conversion coverage script shouldn't have missed this.

tbeswick96:
@dedmen I triple checked there was no implementation for it, but there was an assignment, definition, and declaration for the engine function

jonpas:
GitHub search didn't find it, maybe it's somewhere in the comment?

dedmen:
https://github.com/intercept/intercept/blob/cd971f2e027c9f8bd106fbcdd38a54882c4c5a3d/src/client/intercept/client/sqf/uncategorized.cpp#L218 
Found it. Another Why the F is this on "Don't implement"

jonpas:
This is not under "don't implement", it's under ["investigate"](https://github.com/intercept/intercept/blob/cd971f2e027c9f8bd106fbcdd38a54882c4c5a3d/src/client/intercept/client/sqf/uncategorized.cpp#L160)

PR #152, opened 2017-06-23
Fix some implemetations and add some helper functions

RZSenfo:
I will use this to repair some of the other wrong implemented wrappers too..
(sorry for the commit stuff, this was on done on intercept/develop before, so I needed to create a new branch on my local repo first and arrrgggh)

### DIFF
--- a/src/client/headers/client/sqf/marker.hpp
+++ b/src/client/headers/client/sqf/marker.hpp
@@ -98,8 +98,8 @@ namespace intercept {
         std::vector<location> nearest_locations(const vector3 &pos_, sqf_string_list_const_ref location_types_, float radius_, const vector3 &sort_position_);
         std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_);
         std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_, const vector3 &sort_position_);
-        std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_);
-        std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_, const vector3 &sort_pos_);
+        //std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_);
+        //std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_, const vector3 &sort_pos_);
         std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_, const object &sort_obj_);
         sqf_return_string_list all_variables(const location &value_);
         sqf_return_string name(const location &loc_);
--- a/src/client/headers/client/sqf/marker.hpp
+++ b/src/client/headers/client/sqf/marker.hpp
@@ -96,11 +96,10 @@ namespace intercept {
         vector2 location_position(const location &loc_);
         std::vector<location> nearest_locations(const vector3 &pos_, sqf_string_list_const_ref location_types_, float radius_);
         std::vector<location> nearest_locations(const vector3 &pos_, sqf_string_list_const_ref location_types_, float radius_, const vector3 &sort_position_);
+        std::vector<location> nearest_locations(const vector3 &pos_, sqf_string_list_const_ref location_types_, float radius_, const object &sort_obj_);
         std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_);
         std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_, const vector3 &sort_position_);
-        //std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_);
-        //std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_, const vector3 &sort_pos_);
-        std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_, const object &sort_obj_);
+        std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_, const object &sort_obj_);
         sqf_return_string_list all_variables(const location &value_);
         sqf_return_string name(const location &loc_);
         sqf_return_string type(const location &loc_);
--- a/src/client/headers/client/sqf/marker.hpp
+++ b/src/client/headers/client/sqf/marker.hpp
@@ -98,8 +98,8 @@ namespace intercept {
         std::vector<location> nearest_locations(const vector3 &pos_, sqf_string_list_const_ref location_types_, float radius_, const vector3 &sort_position_);
         std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_);
         std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_, const vector3 &sort_position_);
-        //std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_);
-        //std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_, const vector3 &sort_pos_);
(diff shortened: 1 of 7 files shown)


dedmen:
If you want to merge stuff into develop you should first update your fork ^^

RZSenfo:
But I created the branch as a fork from develop .. however, gonna fix this

dedmen:
Just do another merge from develop into your branch.

jonpas:
Your fork doesn't get automatically updated.

src/client/headers/client/sqf/marker.hpp
```diff
@@ -98,8 +98,8 @@ namespace intercept {
         std::vector<location> nearest_locations(const vector3 &pos_, sqf_string_list_const_ref location_types_, float radius_, const vector3 &sort_position_);
         std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_);
         std::vector<location> nearest_locations(const object &unit_, sqf_string_list_const_ref location_types_, float radius_, const vector3 &sort_position_);
-        std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_);
-        std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_, const vector3 &sort_pos_);
+        //std::vector<location> nearest_locations(const vector3 pos_, sqf_string_list_const_ref loc_types_, float radius_);
```
dedmen: I don't see why you disabled these?

src/client/headers/client/sqf/position.hpp
```diff
@@ -137,7 +137,8 @@ namespace intercept {
         float distance(std::variant<object, vector3> start_, std::variant<object, vector3> end_);
         bool set_vehicle_position(const object &object_, std::variant<std::reference_wrapper<const vector2>, std::reference_wrapper<const vector3>, const object> position_, sqf_string_list_const_ref markers_, float placement_radius_, std::optional<std::string> special_);
         vector3 building_exit(const object &building_, int index_);
-        vector3 building_pos(const object &building_, int index_);
+        vector3 building_pos(const object &building_, unsigned int index_);
```
dedmen: This needs to be made clear to what it does.
You can still call this with -1. Which will result in a huge number when converted to float. Something completly unexpected by the user.
We could just always return a vector of positions. And check for -1 inside the function.

src/client/intercept/client/sqf/common_helpers.cpp
```diff
@@ -155,6 +155,15 @@ namespace intercept {
                 return output;
             }
 
+            std::vector<vector2> __convert_to_vector2_vector(game_value input_) {
+                std::vector<vector2> output;
+                output.reserve(input_.size());
+                for (uint32_t i = 0; i < input_.size(); ++i) {
```
dedmen: I think we should use the new .to_auto_array() func in game_value. And do a foreach loop instead.

src/client/headers/client/sqf/marker.hpp
RZSenfo: Because we have got the same function which takes the pos by ref.
Both don't change anything regarding the position, so why have one fnc which takes ref and one not?
This leads to the problem that the compiler doesn't know which one to choose

src/client/intercept/client/sqf/common_helpers.cpp
RZSenfo: It was just a c&p of convert to vector3 vector, but yeah, will do that

dedmen: Should also do that for the others. You can do that if you want to.

src/client/headers/client/sqf/marker.hpp
dedmen: Ah yeah I see now. Yeah that was definetly an error.
But judging from the Validation error your probably uncommented one nearest_locations too many in the cpp file

src/client/intercept/client/sqf/common_helpers.cpp
RZSenfo: I will have a look as soon as I am at home again

src/client/headers/client/sqf/marker.hpp
dedmen: If you think that's all correct feel free to just remove them instead of uncommenting

dedmen: Yeah I just checked. in cpp you commented 3 implementations of nearest_locations. But in the header you only commented 2

RZSenfo: Ahhh alright, my bad 
I will fix it this evening then.

dedmen:
Mh.. I just wonder why these old commits show up..

RZSenfo:
Me too.
I merged into my fork before creating the PR

src/client/intercept/client/sqf/common_helpers.cpp
RZSenfo: looks like this then, or am I wrong?:
std::vector<vector2> output;
output.reserve(input_.size());
for (auto v2 : input_.to_auto_array()) {

src/client/headers/client/sqf/position.hpp
```diff
@@ -137,7 +137,13 @@ namespace intercept {
         float distance(std::variant<object, vector3> start_, std::variant<object, vector3> end_);
         bool set_vehicle_position(const object &object_, std::variant<std::reference_wrapper<const vector2>, std::reference_wrapper<const vector3>, const object> position_, sqf_string_list_const_ref markers_, float placement_radius_, std::optional<std::string> special_);
         vector3 building_exit(const object &building_, int index_);
-        vector3 building_pos(const object &building_, int index_);
+        
+        /*
```
dedmen: Now you don't need a warning ;) They will see the return type. And the function now does exactly what the Wiki says.

src/client/intercept/client/sqf/common_helpers.cpp
dedmen: `auto& v2`. 
`auto v2` would take a copy of each element.

src/client/headers/client/sqf/position.hpp
RZSenfo: Not exactly, the wiki says either a position or an array of positions

src/client/intercept/client/sqf/common_helpers.cpp
RZSenfo: Alright, also the other helpers always casted to object before pushback, what only worked out because currently you could cast obj to vectors back implicitly

RZSenfo:
Now it's back to the waypoints stuff..
I really don't know where this comes from

src/client/headers/client/sqf/position.hpp
dedmen: But people can see that from function return type. And the function does exactly what you think it does if you look at the function signature and the BIKI.
The other one looked like you could pass -1. Which you could and the compiler wouldn't complain.
But that would actually call buildingPos with a index of `4294967295` instead of `-1` as expected. Which is why that one would have needed documentation and a big WARNING sign.

RZSenfo: I agree on that, yet it is a bit different than the wiki says, so we could just leave that info there and it's all fine

dedmen: Should although conform to the doxygen comment syntax. But we can also do that when we do the general documentation run

dedmen:
The GCC build errors come from develop branch. Nothing for you to worry about.

src/client/intercept/client/sqf/common_helpers.cpp
```diff
@@ -140,26 +140,35 @@ namespace intercept {
...
+                    output.push_back(game_value(gv));
                 }
                 return output;
             }
 
             std::vector<vector3> __convert_to_vector3_vector(game_value input_) {
                 std::vector<vector3> output;
                 output.reserve(input_.size());
-                for (uint32_t i = 0; i < input_.size(); ++i) {
-                    output.push_back(object(input_[i]));
```
dedmen: What? XD It was casting to `object` before? Oh god :D

dedmen:
I'd merge this now then if you're ready.

src/client/intercept/client/sqf/common_helpers.cpp
RZSenfo: Every
Single 
One...
And this was only working because you can imp cast back to vector :D

RZSenfo:
Yeah, go for it.
I will open another one if I do something else

PR #153, opened 2017-07-03
[WIP] Changes allowing intercept to be cross-compiled with Mingw-w64 on linux

overfl0:
32-bit version compiles. I don't know it it works, though :).
~I'm going to check how to compile the 64-bit version tomorrow.~ It compiles now.

If your `i686-w64-mingw32.static.posix-gcc -v` says 7.0.0 or later, you don't need to compile anything. You need the `.posix` version to be installed, otherwise threaded code won't compile.

To compile mingw on an ubuntu machine:

```
apt-get install \
    autoconf automake autopoint bash bison bzip2 flex gettext\
    git g++ gperf intltool libffi-dev libgdk-pixbuf2.0-dev \
    libtool libltdl-dev libssl-dev libxml-parser-perl make \
    openssl p7zip-full patch perl pkg-config python ruby scons \
    sed unzip wget xz-utils\
    g++-multilib libc6-dev-i386\
    libtool-bin # <- Only on 14.10 or later

git clone https://github.com/mxe/mxe
cd mxe
make -j 3 MXE_PLUGIN_DIRS=plugins/gcc7 MXE_TARGETS=i686-w64-mingw32.static.posix gcc
# Grab a cofee because it's going to take ~30 minutes
export PATH=$PATH:"`pwd`/usr/bin"

# Go to intercept repo
i686-w64-mingw32.static.posix-cmake .. && make
```


src/host/common/singleton.hpp
```diff
@@ -5,6 +5,7 @@
 #pragma once
 
 #ifdef __GNUC__
+#undef __forceinline
 #define __forceinline __attribute__((always_inline))
```
overfl0: As I said in the #dev channel: mingw defines this as:
`#define __forceinline inline __attribute__((__always_inline__))`

Is this redefinition correct?

src/host/extensions/search.cpp
```diff
@@ -203,8 +202,8 @@ typedef struct _OBJECT_TYPE_INFORMATION {
     ULONG NonPagedPoolUsage;
 } OBJECT_TYPE_INFORMATION, *POBJECT_TYPE_INFORMATION;
 
-PVOID GetLibraryProcAddress(PSTR LibraryName, PSTR ProcName) {
-    return GetProcAddress(GetModuleHandleA(LibraryName), ProcName);
+PVOID GetLibraryProcAddress(const char* LibraryName, const char* ProcName) {
```
overfl0: Maybe just use LPCSTR here?

src/host/common/singleton.hpp
dedmen: You can just add a additional ifdef for mingw. 
The additional `inline` is not required as that has a different meaning which is mostly unrelated to what __forceinline aims to do.

src/host/extensions/search.cpp
dedmen: I'd prefer to keep as few microsoft-specific types as possible.

src/client/CMakeLists.txt
```diff
@@ -16,7 +16,10 @@ option(USE_STATIC_LINKING "USE_STATIC_LINKING" ON)
 
 
 if(CMAKE_COMPILER_IS_GNUCXX)
-	SET(CMAKE_CXX_FLAGS "-std=c++1z -march=i686 -m32 -O2 -s -fPIC -fpermissive")
+	SET(CMAKE_CXX_FLAGS "-std=c++1z -O2 -s -fPIC -fpermissive")
+	if(NOT USE_64BIT_BUILD)
+		SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -m32 -march=i686")
```
dedmen: Is this really appending? Or does this overwrite?
SET is not the correct way to set these flags anymore. CMake has a addFlag method for that. Dunno what it's called exactly.

src/host/loader/CMakeLists.txt
```diff
@@ -22,6 +22,11 @@ set_target_properties(loader_static PROPERTIES FOLDER intercept)
 if(CMAKE_COMPILER_IS_GNUCXX)
 	set_target_properties(loader_static PROPERTIES LINK_SEARCH_START_STATIC 1)
 	set_target_properties(loader_static PROPERTIES LINK_SEARCH_END_STATIC 1)
-    set_target_properties(loader_static PROPERTIES LINK_SEARCH_START_STATIC 1)
+	set_target_properties(loader_static PROPERTIES LINK_SEARCH_START_STATIC 1)
```
dedmen: We are using tabs in our CMake files apparently. But.. We really should use spaces shouldn't we?

src/host/loader/loader.cpp
```diff
@@ -6,7 +6,7 @@
 #include <dlfcn.h>
 #include <link.h>
 #else
-#include <Psapi.h>
+#include <psapi.h>
 #pragma comment (lib, "Psapi.lib")//GetModuleInformation
```
dedmen: This doesn't work on Mingw? ref: https://github.com/intercept/intercept/pull/153/files#diff-4d906b170cedeeddc09a04d778a79b0aR31
What is with version.lib then?

src/client/CMakeLists.txt
jonpas: It does append, because already set stuff is included as well.

src/host/loader/CMakeLists.txt
jonpas: Yes, space everywhere.

src/host/loader/loader.cpp
overfl0: The included file is lower case. Don't ask me why. That's just the way it is.
For windows it just doesn't matter since the filesystem is case insensitive.

overfl0: What's with version.lib? No idea. What is it for, actually?
The pragma comment stuff is ignored anyway, so I had to add those inside cmake, somewhere.

src/host/loader/CMakeLists.txt
overfl0: Yes, I'm all for spaces! :)
But don't do that within this PR, mkay? Submit your own for that ;P

src/client/CMakeLists.txt
overfl0: The set at line21 does append. You can verify that by doing `make VERBOSE=1`.
I've been using `add_compile_options` but they seemed not to work. Again, don't ask me, I'm not the expert on cmakes. See here:
https://github.com/intercept/intercept/pull/153/commits/ede31ff81d8ef81ea866453b4187c051b08ceb95

src/host/loader/loader.cpp
overfl0: Sigh... sorry, I noticed just now that you were definitely talking about the pragma, not the include change as I thought previously.
Yes, it's just a MSVC thing. https://stackoverflow.com/a/1685261

dedmen: version.lib is for GetFileVersionInfo and such. used in loader.cpp

overfl0: Interesting. It just compiles in MinGW, as you can see on Travis.
The problem with this branch is that it's not yet working perfectly fine - the created dll is somehow faulty and I still have to figure out the cause (that's the reason I've created the other issue for the sanity check).

Maybe it's because it's missing that version.lib? I don't know, but shouldn't it have thrown some kind of linking error if that was really needed for Mingw?

dedmen: Faulty? Explain here or PM me on slack. I'll probably be available in a couple days.

dedmen:
I don't really want this on our travis commit builds. It is taking too long right now anyway.
Release only would be fine though. I guess. But we don't really need it anyway as Appveyor already builds windows binaries.

jonpas:
This takes about the same time as a GCC build, only first build that caches things takes a long time, and that already happened.

overfl0:
As said by @jonpas this takes the same time as a regular GCC build, I would even go farther to say that it's quicker because the setup time is about 1 minute less for mingw (just fetch cross compiler from cache and unzip) than the regular GCC build (install from apt-get).
See here for my today's push to this PR: https://travis-ci.org/intercept/intercept/builds/251698724?utm_source=github_status&utm_medium=notification

The upside of having Mingw builds now is that as soon as someone writes some code that doesn't compile there, the test is going to fail, we won't have to wait until 4am when the runtime tests start and fail (not because Arma but because of our changes in the code). I'll try to finish the preliminary runtime tests in a few days so we can start running those.

If you're really bothered by the one-time (cross compiler) build times, I could prebuild them and store them on a github release (hopefully they will accept a 250MB file) so that travis fetches that instead of building it itself when it's not in the cache. I've kept it the way it is right now because it's more transparent - you don't get a binary blob from a guy that you don't necessarily trust but you can see it's all building from the sources.
But yeah, as i said, if that still bothers you, I can prepare a tbz/package to be fetched.

I'd suggest continuing this cross compiler discussion on slack because this PR is starting to become huge :)

dedmen:
>The upside of having Mingw builds now is that as soon as someone writes some code that doesn't compile there, the test is going to fail
Which our Appveyor build also tells you already.
I also don't see why we need a cross-compiler if we could just fetch the latest build from appveyor and be done with it.

overfl0:
By "there" I mean with MinGW. There are some code quirks and changes that are needed to make it compile with MinGW even though it already compiles on Windows, with AppVeyor. (see: reinterpret_cast changes in search.cpp, for example). I'm not happy about it either but it's not me who makes the rules for compilers :).
If MinGW would be 100% compatible with MSVC, this whole PR wouldn't be needed at all.

I just want to prevent the case where the runtime tests fail not because BI guys changed something but because our changes have made the source not compile with MinGW. And also I would like to stress the need for the possibility to compile it with MinGW instead of just fetching a binary compiled in AppVeyor: you can then get any Intercept version, not just the latest master branch.

dedmen:
You can also get any Intercept version from Appveyor. Every push get's built and archived.

src/client/headers/shared/types.hpp
```diff
@@ -15,6 +15,7 @@
 
 //GNUC somehow can't do it if it's inside shared.hpp.
 #ifdef __GNUC__
+#undef CDECL
```
dedmen: This is wrong. This includes  https://github.com/intercept/intercept/pull/153/files#diff-2238184a23b1b58d9e3a8bbaaffcaf15R11 which does https://github.com/intercept/intercept/pull/153/files#diff-540fab59c55f031c21faa6bdb63b68dbR42 after your change.
meaning trying to #undef twice.

overfl0: @dedmen I'll take a look at your comments and possibly fix my code after the weekend (probably Tuesday)
This PR has been there for quite some time so it's entirely possible that some things have changed in the meantime, meaning the code is no longer correct.

dedmen: This part was not changed since back then. But I can see how you made that mistake. That can easilly happen.. I still wonder why GCC is too dumb...

overfl0: No, wait! I've read that again and it looks correct to me:
It includes `shared.hpp`, which does indeed `#undef CDECL` but it defines that just afterwards.
This means that If you want to `#define CDECL` here, in line 19, you will get the error that it's already defined (which was done in `shared.hpp`) hence the `#undef`.

Maybe what's really needed is a
`#ifdef, #undef, #endif` sequence?

dedmen: Well... you need the #undef to define it already as the same thing it has been defined before.. 
Which is so dumb. I could get GCC to recognize the #define from shared.hpp. What I would prefer is compeltly removing that block from types.hpp

overfl0:
I guess we can remove this one, eh?
Since Intercept has been using Github Actions that normally build everything for both OS-es we don't have any need for cross-compiling. Not that this code has been updated in the last 4 years :P

PR #157, opened 2017-07-17
Fixes

RZSenfo:


src/client/intercept/client/sqf/core.cpp
```diff
@@ -1,4 +1,4 @@
-#include "chat.hpp"
```
dedmen: I'm quite sure I fixed the first line in all sqf files. So this probably changes it to be wrong. UTF-8 with or without BOM. And this is using the wrong variant.

RZSenfo: ummm, Idk why this is even in there

dedmen: VS saved the file probably with UTF-8 with BOM. We use without BOM

dedmen: Editorconfig https://github.com/intercept/intercept/blob/master/.editorconfig#L6 States utf-8 without BOM though. So your VS is probably ignoring it

RZSenfo: ok will check that

jonpas: Editorconfig integrated into VS2017 has a [bug](https://developercommunity.visualstudio.com/content/problem/22922/editorconfig-support-interprets-charset-utf-8-as-u.html) which they apparently fixed months ago but still hasn't been released. Just don't use it, it simply doesn't work yet.

dedmen: That's why I changed that in every of the sqf files. Because I was using the dev preview of VS that probably already has the fix

jonpas: So it will finally soon be released!?

src/client/headers/client/sqf/units.hpp
```diff
@@ -111,7 +111,7 @@ namespace intercept {
         rv_vehicle_role assigned_vehicle_role(const object &unit_);
         group get_group(const object &unit_);  // originally "group", but is already a type
         std::vector<object> group_selected_units(const object &unit_);
-        sqf_return_string_list squad_params(const object &unit_);
+        std::vector<sqf_return_string_list> squad_params(const object &unit_);
```
dedmen: Are you sure? A list of string lists?

src/client/intercept/client/sqf/ai.cpp
```diff
@@ -1,4 +1,4 @@
-#include "ai.hpp"
```
dedmen: Nooo :/

src/client/intercept/client/sqf/ai.cpp
```diff
@@ -644,11 +644,16 @@ namespace intercept {
...
 
-        void set_combat_mode(std::variant<group, object> group_, sqf_string_const_ref mode_) {
-            if (group_.index() == 0)
+        void set_combat_mode(std::variant<group, object, team_member> group_, sqf_string_const_ref mode_) {
+            if (group_.index() == 0) {
                 host::functions.invoke_raw_binary(__sqf::binary__setcombatmode__object_group__string__ret__nothing, std::get<0>(group_), mode_);
-            return;
-            host::functions.invoke_raw_binary(__sqf::binary__setcombatmode__object_group__string__ret__nothing, std::get<1>(group_), mode_);
+            }
+            else if (group_.index() == 0) {
```
dedmen: if index==0 else if index==0 else?
This doesn't make sense

src/client/intercept/client/sqf/curator.cpp
```diff
@@ -1,4 +1,4 @@
-#include "curator.hpp"
```
dedmen: :/ Open it in Notepad++ and convert it to UTF-8 without BOM and commit again

src/client/intercept/client/sqf/curator.cpp
```diff
@@ -13,8 +13,24 @@ namespace intercept {
             return __helpers::__retrieve_nular_object(__sqf::nular__curatormouseover__ret__object);
         }
 
-        std::vector<object> curator_selected() {
-            return __helpers::__convert_to_objects_vector(host::functions.invoke_raw_nular(__sqf::nular__curatorselected__ret__array));
+        _curator_selected_return curator_selected() {
```
dedmen: datatypes don't start with an _

src/client/intercept/client/sqf/units.cpp
```diff
@@ -377,8 +377,13 @@ namespace intercept {
             return __helpers::__convert_to_objects_vector(host::functions.invoke_raw_unary(__sqf::unary__groupselectedunits__object__ret__array, unit_));
         }
 
-        sqf_return_string_list squad_params(const object &unit_) {
-            return __helpers::__convert_to_strings_vector(host::functions.invoke_raw_unary(__sqf::unary__squadparams__object__ret__array, unit_));
+        std::vector<sqf_return_string_list> squad_params(const object &unit_) {
+            std::vector<sqf_return_string_list> _temp;
+            game_value _engine_result = host::functions.invoke_raw_unary(__sqf::unary__squadparams__object__ret__array, unit_);
+            for (unsigned int _k = 0; _k < _engine_result.size(); _k++) {
```
dedmen: This is not how you traverse a array in a game_value

src/client/intercept/client/sqf/units.cpp
```diff
@@ -377,8 +377,13 @@ namespace intercept {
             return __helpers::__convert_to_objects_vector(host::functions.invoke_raw_unary(__sqf::unary__groupselectedunits__object__ret__array, unit_));
         }
 
-        sqf_return_string_list squad_params(const object &unit_) {
-            return __helpers::__convert_to_strings_vector(host::functions.invoke_raw_unary(__sqf::unary__squadparams__object__ret__array, unit_));
+        std::vector<sqf_return_string_list> squad_params(const object &unit_) {
+            std::vector<sqf_return_string_list> _temp;
```
dedmen: you are never calling reserve

RZSenfo: why should I ?

src/client/headers/client/sqf/units.hpp
RZSenfo: thats what the report and the wiki says

src/client/intercept/client/sqf/units.cpp
RZSenfo: how do we?

src/client/intercept/client/sqf/units.cpp
dedmen: Because you know how many elements you want to insert. And pushback will do many reallocations if you don't.

src/client/intercept/client/sqf/curator.cpp
```diff
@@ -13,18 +13,16 @@ namespace intercept {
...
+            _wps.reserve(_raw[2].size());
+            for (auto &gv : _raw[2].to_array()) {
+                _wps.push_back(game_value(gv));
             }
 
-            _curator_selected_return _return = _curator_selected_return(
+            curator_selected_return _return = curator_selected_return(
                 __helpers::__convert_to_objects_vector(_raw[0]),
                 __helpers::__convert_to_groups_vector(_raw[1]),
                 _wps,
```
dedmen: std::move(wps) to prevent copy

src/client/intercept/client/sqf/curator.cpp
```diff
@@ -13,18 +13,16 @@ namespace intercept {
...
             std::vector<game_value> _raw = __helpers::__convert_to_game_value_vector(host::functions.invoke_raw_nular(__sqf::nular__curatorselected__ret__array));
             
-
-            std::vector<game_value> _wptemp = __helpers::__convert_to_game_value_vector(_raw[2]);
             std::vector<intercept::sqf::waypoint> _wps;
-
-            for (auto& _wp : _wptemp) {
-                _wps.push_back(std::move(intercept::sqf::waypoint(_wp)));
+            _wps.reserve(_raw[2].size());
+            for (auto &gv : _raw[2].to_array()) {
```
dedmen: you might be able to do `_wps.insert(_wps.end(), _raw[2].to_array().begin() ,_raw[2].to_array().end())` here and remove the need for the for loop and reserve

src/client/headers/client/sqf/ai.hpp
```diff
@@ -1,4 +1,4 @@
-/*!
```
dedmen: encoding

src/client/intercept/client/sqf/units.cpp
RZSenfo: Ahh ok, yeah, already figured so

src/client/intercept/client/sqf/curator.cpp
RZSenfo: will test that

src/client/headers/client/sqf/ai.hpp
RZSenfo: this really drives me crazy.. everytime I open the file, it gets reset to "with BOM"

jonpas: Disable EditorConfig integration in VS settings.

src/client/intercept/client/sqf/curator.cpp
RZSenfo: shouldn't that to_array result be saved to prevent it from being created twice?

dedmen: it's just a pointer cast internally afaik.

RZSenfo: doesnt work, since we need to convert to waypoint 
for game_value it would work

dedmen: okey yeah. Oopsie

RZSenfo: no problemo

src/client/headers/client/sqf/units.hpp
dedmen: It's not. It's 2 arrays of strings. Should be 2 structs at best. Still gonna merge this. you can fix that later if you want.

PR #162, opened 2017-09-11
Fix unitloadout crash on null objects

RZSenfo:


src/client/headers/client/sqf/inventory.hpp
```diff
@@ -357,13 +357,17 @@ namespace intercept {
...
-            std::string silencer = "";
+            std::string silencer;
             std::string laser; 
             std::string optics;
             rv_magazine_info primary_muzzle_magazine;
             rv_magazine_info secondary_muzzle_magazine;
             std::string bipod;
 
+            rv_weapon_info() {
+                weapon = "";
```
dedmen: weapon is already empty string. There is no need for this.

src/client/headers/client/sqf/inventory.hpp
```diff
@@ -397,6 +401,10 @@ namespace intercept {
             std::string container;
             std::vector<rv_magazine_info> items;
 
+            rv_container_info() {
+                container = "";
```
dedmen: same. No need to initialize something to something it already is.

src/client/headers/client/sqf/inventory.hpp
```diff
@@ -431,17 +439,20 @@ namespace intercept {
...
-                                                                 uniform(ret_game_value_[3]),
-                                                                 vest(ret_game_value_[4]),
-                                                                 backpack(ret_game_value_[5]),
-                                                                 headgear(ret_game_value_[6]),
-                                                                 facewear(ret_game_value_[7]),
-                                                                 binocular(ret_game_value_[8]) {
-                auto_array<game_value> _tmp = ret_game_value_[9].to_array();
-                assigned_items = std::vector<std::string>(_tmp.begin(), _tmp.end());
+            rv_unit_loadout(const game_value &ret_game_value_) {
+                if (ret_game_value_.size() > 0) {
```
dedmen: I'd prefer a `== 0) return;` over a whole lot of indentation
Also.. Does this return an empty array (I know you told me in Slack but I forgot ._.) ? If it is not an array then `size` will throw an exception. I don't know if isNull check might work.

src/client/headers/client/sqf/inventory.hpp
```diff
@@ -431,17 +439,20 @@ namespace intercept {
...
+                    primary = ret_game_value_[0];
+                    secondary = ret_game_value_[1];
+                    handgun = ret_game_value_[2];
+                    uniform = ret_game_value_[3];
+                    vest = ret_game_value_[4];
+                    backpack = ret_game_value_[5];
+                    headgear = static_cast<std::string>(ret_game_value_[6]);
+                    facewear = static_cast<std::string>(ret_game_value_[7]);
+                    binocular = ret_game_value_[8];
+                    auto_array<game_value> _tmp = ret_game_value_[9].to_array();
```
dedmen: use `auto& _tmp =` Here you are copying the whole array.

PR #164, opened 2017-09-15
Added spawn and isEqual* Wrappers

dedmen:
- binary__isequalto__any__any__ret__bool
- binary__isequaltype__any__any__ret__bool
- binary__isequaltypeall__array__any__ret__bool
- binary__isequaltypeany__any__array__ret__bool
- binary__isequaltypearray__array__array__ret__bool
- binary__isequaltypeparams__any__array__ret__bool
- binary__spawn__any__code__ret__script

I decided to implement the isEqual* wrappers because they are very useful and should have no huge performance impact. All they are doing in engine side is basically comparing numbers.
We can already do isEqualTo and isEqualType using our methods. But they are clunky and in the end do the same as the in-engine solutions just more complicated.

@RZSenfo requested `spawn` and I see this being useful to easilly launch tasks you don't care about when they complete.

### DIFF
--- a/src/client/headers/client/sqf/core.hpp
+++ b/src/client/headers/client/sqf/core.hpp
@@ -1,4 +1,4 @@
-﻿/*!
+/*!
 @file
 @author Verox (verox.averre@gmail.com)
 @author Nou (korewananda@gmail.com)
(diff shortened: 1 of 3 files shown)


dedmen:
@RZSenfo Did you test the isEqualTo, spawn and maybe isEqualType?

dedmen:
Let's just assume they work :D

PR #169, opened 2017-10-31
Develop -> master

dedmen:
- Add invoker_lock::unlock 
- 1.70->1.76 SQF pointer list updates. Some wrappers still missing.
- Add INTERCEPT_SQF_STRTYPE_RSTRING define and CMake option to make Intercept use engine types instead of converting from/to std::string. You need to use the same setting for intercept_client.lib as you use in your plugin.
- Add `game_value::type_enum()` which returns the type as enum.
- Added `rv_turret_path type` and use it instead of using vector<int> or vector<float> or vector<string> or whatever else was used because people didn't know what to use. **Plugin breaking if you use Eventhandlers or Functions with turret paths.**
- Added r_string `compare_case_insensitive()` and `empty()` and `size()` which is just redirect to `length()`.
- Lots of refactoring as always.
- Added r_string `operator!=` `operator<` `operator+` `append` `to_lower` `begin()` `end()` `front()`
- Added lots of Doxygen documentation in code.
- Added `game_value::is_null()` which should work just like SQF `isNull`
- Added `client::generate_custom_callback()` Which generates a small SQF script that calls directly into a passed lambda function.
- Added post_start EH (XEH_preStart) because you couldn't setVariable to uiNamespace at preStart.
- Added vector `magnitude_squared` function
- Added auto_array range erase function. Meaning you can now use `Array.erase(std::unique(Array.begin(), Array.end()), Array.end());` which is the Intercept equivalent of SQF's `Array arrayIntersect Array`.
- Changed get_unit_loadout to return raw game_value because there might be a lot of cases where one just wants to store the data and pass it along but not parse it.
- Fixed wrong implementation of HitPart eventhandlers arguments.


dedmen:
I wan't to release my first Intercept Mod and I don't want to tell people to use develop branch.
I need dis @jonpas

PR #171, opened 2017-11-11
x64 binary hotfix

dedmen:
This fixes Intercept not being able to find the game state pointer on latest x64 binaries.
Intercept can now automatically search and find the game state. No need to fix these offsets anymore.


dedmen:
There is one more change coming which will add signing for the core binaries.

PR #172, opened 2017-11-12
Fix linux library errors

dedmen:
Will need binary rebuild because the gameState finder algo was broken on 32-bit.
I tested this on Linux server. and Arma stable 32/64 bit.


dedmen:
libstdc++ is still needed. And still in arma3 main directory

[linuxbuild.zip](https://github.com/intercept/intercept/files/1465024/P.zip)

PR #176, opened 2017-11-15
Replace easylogging with spdlog

dedmen:
spdlog is still bigger than what we need. But atleast it doesn't crash instantly on Linux. Perf and resulting binary size is also better.


dedmen:
Tested together with #175 on Linux. Working just fine.
@jonpas

Issue #138, opened 2017-05-29
Incorrect Wrapper Implementations


SilentSpike:
Generic Issue to list any incorrect wrapper implementations. Currently known:

- [ ] `curator_selected()` returns `std::vector<object>` but the actual sqf command returns an array of the form: `[[objects],[groups],[waypoints],[markers]]`

jonpas:
Added:
- `get/set_unit_loadout()` using wrong data

dedmen:
wrong data? Sure that's not already fixed?

jonpas:
Not if you fixed it, but I haven't checked yet.

dedmen:
`getUnitLoadout config` is missing
Also https://community.bistudio.com/wiki/setSkill_array is missing.

dedmen:
setVariable public variants of many setVariable wrappers.
Already added object's one because I needed it

dedmen:
set_group_id params are wrong
https://community.bistudio.com/wiki/setGroupId see example 1.
We onle need params if nameFormat has %1 like formatting.

dedmen:
`float count(const config &config_entry_);` should return int or size_t

dedmen:
![](https://dedmen.de/sharex/94I6dcpl.png)

I think the lbSetPicture thingies are on "Don't Implement" list because of missing documentation. They have documenation now.

RZSenfo:
sqf::building_pos has no variant that returns the array of all positions, when -1 is the index
sqf::in_area_array ... the params.. dunno how this is supposed to work
sqf::nearest_locations multiple instances, because subtypes can convert to each other

dedmen:
sqf::world_to_screen don't take in_screen as parameter but return pair<bool,pos> instead. Or pair<pos,bool> ? Second would match std::map::insert return value which is pair<iterator,bool>

dedmen:
sqf::world_to_screen will now return `std::optional<vector2>` instead of a on_screen variable as reference.
This is the cleanest way IMO

dedmen:
Everything besides 
https://github.com/intercept/intercept/issues/138#issuecomment-306857557
https://github.com/intercept/intercept/issues/138#issuecomment-309159809
https://github.com/intercept/intercept/issues/138#issuecomment-307378238
and `curator_selected()` is fixed in https://github.com/intercept/intercept/pull/178

dedmen:
intercept::sqf::agents(); returns std::vector<intercept::types::team_member>

[...]

PR #178, opened 2017-11-18
Wrapper cleanup #138

dedmen:


dedmen:
* Remove redundant is_equal_to
* setVariable with public flags
* world_to_screen return optional instead of taking on_screen argument as reference
* added side_get and group_get alternatives to get_side/get_group
* squad_params use typedef
* Implement some new wrappers
* addforce/addtorque
* backpack_space_for
* vectorModel commands
* getMousePosition isuiContext getMissionVersion

PR #183, opened 2017-11-29
client_eventhandlers_clear linker fix for x86

chris579:
Fixes `error LNK2001: unresolved external symbol client_eventhandlers_clear` for x86

### DIFF
--- a/src/client/headers/client/eventhandlers.hpp
+++ b/src/client/headers/client/eventhandlers.hpp
@@ -10,7 +10,13 @@ https://github.com/NouberNou/intercept
 Provides the ability to assign Eventhandlers without going through SQF
 */
 #pragma once
+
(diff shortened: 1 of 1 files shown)


dedmen:
Tested I assume?
I also fiddled with this and my solution also has the function declarations moved into the header so they are publicly visible.

chris579:
> Tested I assume?

On my local machine this solution works and could act as a temp fix until you push your solution. Alternatively you could push your solution and I close this pr.

dedmen:
If yours works it's better than mine

chris579:
You should give this a test in one of your projects. I know I'm developing in a different way because I'm not using CMAKE and so on. But in theory it should work.

dedmen:
My test on 32bit works. So..

PR #184, opened 2017-11-29
CI build fix attempt

chris579:


dedmen:
Is this still current? I assume it isn't :D

PR #185, opened 2017-11-30
Ref swap for faster r_string move construct

dedmen:
Current r_string move constructor still increases/modifies the refcount.
This add's a swap function that doesn't modify refcount. Thus faster


src/client/headers/shared/types.hpp
```diff
@@ -240,6 +240,12 @@ namespace intercept {
                 return *this;
             }
 
+            void swap(ref &other) {
```
dedmen: Code conventions you idiot?! Leading `_` ?!

PR #198, opened 2018-02-20
Rename exported functions according to convention

dedmen:
Also Register Function -> Register Command. Because Function != Command. And Intercept-CBA implements proper functions.


dedmen:
Some things take their time :D

PR #199, opened 2018-03-07
fix arguments mixup for sqf displayctrl command

brainslush:


dedmen:
Rdy? We'll need to remove the const& for the display parameter before helper classes arrive. But there's still some:tm: time till then.

PR #200, opened 2018-03-13
Use crossplatform way of comparing r_string

dedmen:


dedmen:
Tested on Linux already together with intercept_cba and my script profiler. Linux test missing was the main concern for not merging this earlier.

PR #202, opened 2018-05-30
Make Variable namespaces accessible

dedmen:
Another huge PR ^^

- Allow to directly set local variables to the local script's scope which is being used for sqf::call
- Fix mapStringToClass not setting existing items on insert
- add i_ref (refcount wrapper for IRef's used by Debugger Interfaces)
- add r_string::substr which returns string_view
- add map_string_to_class::remove
- Remove windows.h include from plugin header files
- Add stack_array (auto_array with push/pop functions)
- Fixed ignored code signing cert failures
- Fix number sign missmatch warning when compiling containers.hpp

Ported stuff from ArmaDebugEngine
- Implement I_debug_value (Mostly useless. But we need it implemented to keep correct vtable inheritance)
- Correctly implement game_data_namespace (Old one was just a placeholder)

### DIFF
--- a/src/client/headers/shared.hpp
+++ b/src/client/headers/shared.hpp
@@ -1,15 +1,10 @@
-﻿#pragma once
+#pragma once
 
 // This is a warning normally for returning references to local/stack allocated variables
 // It is so dangerous though that we need to force it as a warning because it can break
 // everything.
 #pragma warning( error : 4172 )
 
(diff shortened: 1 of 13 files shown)


dedmen:
Ran these changes for the last 5 days. No bugs found in the new stuff. I'd say this is good.

PR #206, opened 2018-07-17
particle_array and particle_shape correction

LtJulius:
corrected wrong binary function pointer
corrected wrong attribute of rv_particle_array
corrected wrong game_value cast of rv_particle_shape ( causing <no-shape> )

### DIFF
--- a/src/client/headers/shared/client_types.hpp
+++ b/src/client/headers/shared/client_types.hpp
@@ -152,7 +152,7 @@ class internal_object : public game_value {
 
     struct rv_particle_shape {
         std::string file;
-        int ntieth = 16; //this param is a const internally
+        int ntieth;
         int row;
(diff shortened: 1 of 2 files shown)


src/client/headers/shared/client_types.hpp
```diff
@@ -152,15 +152,15 @@ class internal_object : public game_value {
 
     struct rv_particle_shape {
         std::string file;
-        int ntieth = 16; //this param is a const internally
+        int ntieth;
         int row;
```
dedmen: does row/column make sense here? Wiki says particleFSIndex and particleFSFrameCount which are very different things.
https://community.bistudio.com/wiki/setParticleParams
Maybe change that too? Just use the names from wiki

PR #209, opened 2018-08-07
Fix setGroupID

dedmen:
We were passing
`group setGroupId [nameFormat, [nameParam1, ..., nameParamN] ]`
instead of
`group setGroupId [nameFormat, nameParam1, ..., nameParamN]`

Also added a little bit of auto_array documentation and a convenience insert range function.


dedmen:
https://github.com/intercept/intercept/blob/master/src/client/headers/client/sqf/sqf.hpp#L3 actions.hpp missing

PR #215, opened 2019-01-17
Add support for building with Clang

Dahlgren:
Useful for local development and unit testing.
Probably not usable together with Arma.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,9 +26,10 @@ matrix:
         fi
 
     ###############################################################################
-    # Job #2 - Building intercept
+    # Job #2 - Building intercept with GCC
     ###############################################################################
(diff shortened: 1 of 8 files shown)


dedmen:
Is it possible to show some kind of warning that the built library will probably not work with Arma?
I mean.. I didn't test yet, but I kinda doubt it's binary compatible.

Dahlgren:
Sure, there's `#pragma message "some diagnostic message"`. I'm not sure it can be seen with all the current warnings triggered when building with clang. Any particular file you want it added to?

Dahlgren:
@dedmen added clang `clang is unsupported for usage with Arma` message to both CMake setup and `shared.hpp` for host library

PR #219, opened 2019-03-14
Fix broken GNU build due to erroneous character in CMake config

Dahlgren:


dedmen:
Manually pulled RelWithDeb info out of there. I made that mistake in other places too but always noticed it myself

PR #221, opened 2019-03-14
Set 32 bit as arch for all host directories when using Clang or GCC

Dahlgren:


dedmen:
Shouldn't it be enough to just set it on the target?
Which is already done here:
https://github.com/intercept/intercept/blob/3487c762803b79e14fc5a575a61368cf519584d6/src/host/intercept_dll/CMakeLists.txt#L39




Intercept_client library wasn't converted yet
https://github.com/intercept/intercept/blob/3487c762803b79e14fc5a575a61368cf519584d6/src/client/CMakeLists.txt#L14
But I am aiming to get rid of that anyway.

Dahlgren:
No, as can be seen on the current [master branch](https://travis-ci.org/intercept/intercept/jobs/506465315#L4273) the linker fails on linux due to the other host libraries being compiled as x86-64

```
/usr/bin/ld: i386:x86-64 architecture of input file `../common/libcommon_static.a(shared.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../loader/loader_static.a(loader.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../loader/loader_static.a(containers.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../loader/loader_static.a(types.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../controller/controller_static.a(controller.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../invoker/invoker_static.a(invoker.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../invoker/invoker_static.a(sqf_functions.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../invoker/invoker_static.a(eventhandlers.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../extensions/extensions_static.a(extensions.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../extensions/extensions_static.a(search.cpp.o)' is incompatible with i386 output
/usr/bin/ld: i386:x86-64 architecture of input file `../extensions/extensions_static.a(export.cpp.o)' is incompatible with i386 output
```

dedmen:
Ah yeah https://github.com/intercept/intercept/blob/master/src/host/loader/CMakeLists.txt#L11
I feel like it would be best to make a cmake function to set the compile options. And call it on every output

PR #222, opened 2019-03-15
Ensure pthread is linked when using Clang

Dahlgren:


Dahlgren:
Not sure if this is the best solution but now both GCC and Clang works fine and Travis CI is green

src/host/intercept_dll/CMakeLists.txt
```diff
@@ -50,6 +50,7 @@ if((CMAKE_CXX_COMPILER_ID MATCHES "Clang") OR (CMAKE_CXX_COMPILER_ID MATCHES "GN
 	set(CMAKE_FIND_LIBRARY_SUFFIXES ".a")
 
 	target_link_libraries(${INTERCEPT_EXTENSION_NAME} "libstdc++.a")
+	target_link_libraries(${INTERCEPT_EXTENSION_NAME} "pthread")
```
dedmen: target_link_libraries (${INTERCEPT_PLUGIN_NAME} ${CMAKE_THREAD_LIBS_INIT})
```suggestion
	find_package (Threads)
	target_link_libraries (${INTERCEPT_EXTENSION_NAME} ${CMAKE_THREAD_LIBS_INIT})
```
the find_package actually belongs on a higher level. But shouldn't hurt.

Dahlgren: Updated with suggested change

PR #223, opened 2019-04-13
Corrected & Updated SQF Commands

LtJulius:
Added:
+ synchonized_triggers for waypoints

Fixed:
^ signature of intercept command ( result to intersect_result_list )


src/client/headers/client/sqf/intersects.hpp
```diff
@@ -1,4 +1,4 @@
-﻿/*!
+/*!
```
dedmen: should be UTF-8 without BOM

src/client/headers/client/sqf/intersects.hpp
```diff
@@ -26,14 +26,22 @@ namespace intercept {
         };
         typedef std::vector<intersect_surfaces> intersect_surfaces_list;
 
+        struct intersect_result {
+            std::string selection_name;
```
dedmen: use intercept string type. Don't know what the exact name is. The thing that other functions that return strings already return

src/client/intercept/client/sqf/intersects.cpp
```diff
@@ -5,6 +5,23 @@
 namespace intercept {
     namespace sqf {
         namespace __helpers {
+            intersect_result_list __intersects(const game_value &intersects_value_) {
+                intersect_result_list output;
+                output.reserve(intersects_value_.size());
+
+                for (uint32_t i = 0; i < intersects_value_.size(); ++i) {
```
dedmen: for (auto& element : intersects_value_.as_array())

src/client/headers/client/sqf/intersects.hpp
dedmen: It is :+1:

PR #228, opened 2019-05-07
Removed inline from lock/unlock

TheWillard:
Removed `inline` from `lock()`/`unlock()` because it might get optimized away as discussed on Slack.


dedmen:
Caused undefined symbol error on unlock because unlock is not referenced in cpp file. Other fix could be calling unlock from destructor but might harm perf by a nanosecond or so.

PR #229, opened 2019-07-07
Updated params for nearestTerrainObjects to support changes introduce…

Lifetap:
…d in Arma 3 v1.61.136035 and Arma 3 v1.65.138818


dedmen:
Why add 2 new variants with sort/mode parameter, if sort parameter is default true and mode default false.
Just add the parameters to the old 2 variants, and them a default value.

Lifetap:
Code updated to replace overloading with default params

PR #230, opened 2019-07-08
Call Refactor

LtJulius:


src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+    }  // namespace sqf
+}  // namespace intercept
+
+#include "client/pointers.hpp"
+
+namespace intercept {
+    namespace sqf {
+
+        /**
+         * We need other functions do be defined in header before we can template call functions
```
dedmen: yeah I don't like that, but I guess every other way is simply too ugly.

src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+         *  Otherwise we need forward declaration
+         */
+        template<typename T>
+        T call2(const code &code_, game_value args_) {
+            game_value args({ args_, code_ });
+
+            sqf::set_variable(sqf::mission_namespace(), "INTERCEPT_CALL_ARGS"sv, args);
+
+            code wrapper = sqf::get_variable(sqf::mission_namespace(), "intercept_fnc_isNilWrapper"sv);
+            if (wrapper.is_nil()) {
```
dedmen: I think.. We can fix this by compileFinal'ing into uiNamespace at preStart? At early preStart before Intercept gets initialized to make sure it happens before everyone else.
I fell into the trap in the past where the sqf was updated but the cpp copy of it wasn't.

dedmen: I somehow like the idea of using parsingNamespace :D Just because noone uses it ever.

src/client/headers/client/sqf/core.hpp
```diff
@@ -25,12 +25,9 @@ namespace intercept {
         ///Use this to check if fastCall is available. Which calls the engine function directly instead of going through a isNil wrapper
         bool _has_fast_call();
         //This variant is special in that it works before preInit. The one without args doesn't
-        game_value call(const code &code_, game_value args_);
-        game_value call(const code &code_);
-        game_value call2(const code &code_, game_value args_);
-        game_value call2(const code &code_);
```
dedmen: maybe we need to keep these, with a [deprecated] attribute

src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+
+        template<typename T>
+        T call2(const code& code_)
+        {
+            return call2<T>(code_, game_value());
+        }
+
+        /**
+         * Special treatment for void type
+         *  Void inherent modifies and uses other wrapper to avoid fetching return values
```
dedmen: >Void inherent modifies

That seems unclear

src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+                sqf::set_variable(sqf::mission_namespace(), "intercept_fnc_voidWrapper"sv, static_cast<game_value>(wrapper));
+            }
+
+            host::functions.invoke_raw_unary(
+                __sqf::unary__isnil__code_string__ret__bool,
+                wrapper);
+        }
+
+        template<>
+        inline void call2<void>(const code& code_)
```
dedmen: do we really need this? the un-specialized wrapper should just foward the void to the call2 with args, which should call the specialized version

src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+            gs->set_local_variable(arname, std::move(args_));
+            gs->set_local_variable(codename, code_);
+
+            auto ret = ef(*data, ns, fname);
+            return static_cast<T>(ret);
+        }
+
+        template<typename T>
+        T call(const code& code_)
+        {
```
dedmen: Doesn't conform to code style. braces on same line with space before it.

src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+        inline void call<void>(const code& code_, game_value args_)
+        {
+            auto allo = host::functions.get_engine_allocator();
+            auto ef = allo->evaluate_func;
+            auto gs = allo->gameState;
+
+            if (!_has_fast_call())
+            {
+                call2<void>(code_, std::move(args_));
+                return;
```
dedmen: we should be able to just `return call2<void> ...` right? returning a void is valid afaik.

src/client/headers/client/sqf/vehicles.hpp
```diff
@@ -101,7 +101,7 @@ namespace intercept {
         object assigned_gunner(const object &veh_);
         object assigned_target(const object &veh_);
         object commander(const object &veh_);
-        void create_vehicle_crew(const object &veh_);
+        group create_vehicle_crew(const object &veh_);
```
dedmen: wat?

src/client/headers/client/sqf_pointers_declaration.hpp
```diff
@@ -1,4 +1,4 @@
-//Exported Pointer Definitions For: Arma 3 188.145302 - Profile
+//Exported Pointer Definitions For: Arma 3 192.145649 - Profile
```
dedmen: don't pull other unrelated pull requests into this.

src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+#include "client/pointers.hpp"
+
+namespace intercept {
+    namespace sqf {
+
+        /**
+         * We need other functions do be defined in header before we can template call functions
+         *  Otherwise we need forward declaration
+         */
+        template<typename T>
```
dedmen:
```suggestion
        template<typename T = game_value>
```
such that we keep the old default behavior without having to update codebase

src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+        T call2(const code &code_, game_value args_) {
```
dedmen: I'd like to turn call2 private, or move it to internal namespace, people shouldn't touch this directly I think.
But we can do that in seperate PR.

src/client/headers/client/sqf/core.hpp
```diff
@@ -234,10 +231,172 @@ namespace intercept {
...
+                    "\
+                    (missionNamespace getVariable \"INTERCEPT_CALL_ARGS\") params[\"_args\", \"_code\"];\
+                    missionNamespace setVariable[\"INTERCEPT_CALL_RETURN\", if (isNil \"_args\") then {call _code} else {_args call _code}];\
+                    "sv);
+                sqf::set_variable(sqf::mission_namespace(), "intercept_fnc_isNilWrapper"sv, static_cast<game_value>(wrapper));
+            }
+            host::functions.invoke_raw_unary(
+                __sqf::unary__isnil__code_string__ret__bool,
+                wrapper);
+            return static_cast<T>(sqf::get_variable(sqf::mission_namespace(), "INTERCEPT_CALL_RETURN"sv));
```
dedmen: We need to make sure that T is actually really convertible from game_value.. 
Though actually, do we need to? It probably throws compile error if you can't convert?

src/client/intercept/client/sqf/core.cpp
```diff
@@ -18,112 +18,26 @@ namespace intercept {
...
-
         script spawn(game_value args, const code& code_) {
             return host::functions.invoke_raw_binary(
                 __sqf::binary__spawn__any__code__ret__script,
                 args,
                 code_);
         }
 
+        game_value call_raw(const code &code_, game_value args_) {
+            return host::functions.invoke_raw_binary(
```
dedmen: We previously didn't expose it, because it simply didn't work as people expected it to work. So not sure if we want it now.
We could if we add documentation to it and clearly explain what's wrong with it.

dedmen:
?

src/client/headers/client/sqf/sqf.hpp
```diff
@@ -55,9 +55,9 @@ class __inline_sqf_helper_launcher {
             _sqf = std::regex_replace(_sqf, escape, "$1");
             const auto sqf_fnc = intercept::sqf::compile(_sqf);
             if (_capture_return) {
-                *_capture_return = intercept::sqf::call(sqf_fnc, _capture_args);
+                *_capture_return = intercept::sqf::call<game_value>(sqf_fnc, _capture_args);
```
dedmen: This shouldn't be needed anymore right?
```suggestion
                *_capture_return = intercept::sqf::call(sqf_fnc, _capture_args);
```

PR #236, opened 2020-04-20
Fix Travis CI builds

Dahlgren:


Dahlgren:
https://travis-ci.org/github/intercept/intercept/builds/677071561

Seems like Travis CI status reporting is not working as it should, perhaps the integration token must be updated?

PR #237, opened 2020-05-19
Add logs directory if needed

Dahlgren:
Fixes crash on Linux
Fixes missing logs on Windows


Dahlgren:
Tested successfully on Windows. Got a segfault on Linux but might be unrelated, needs some more testing.

src/host/intercept_dll/intercept_dll.cpp
```diff
@@ -119,12 +120,20 @@ __attribute__((constructor))
             size_t path_end_ = arg_line.find("\"", path_start_); // find ending quotationmark
 
             if (path_start_ == std::string::npos) {
+                if (!std::filesystem::is_directory("logs") || !std::filesystem::exists("logs")) {
```
jokoho48, line 123: (!A || !B) == !(A && B)

src/host/intercept_dll/intercept_dll.cpp
```diff
@@ -119,12 +120,20 @@ __attribute__((constructor))
...
+                    std::filesystem::create_directory("logs");
+                }
+
                 logging::logfile = spdlog::rotating_logger_mt("logfile", "logs/intercept_dll.log", 1024000, 3);
             } else {
                 std::string _path = arg_line.substr(path_start_, path_end_ - path_start_);
                 logging::logfile = spdlog::rotating_logger_mt("logfile", _path, 1024000, 3);
             }
         } else {
+            if (!std::filesystem::is_directory("logs") || !std::filesystem::exists("logs")) {
```
jokoho48, line 133: Same as above

dedmen:
I intentionally didn't auto create, because I didn't want people to be spammed with logs if they don't need them 🙃

src/host/intercept_dll/intercept_dll.cpp
dedmen: Yes. Correct. But what do you mean with that?
the && would be less efficient if only the first is true so this is fine.
But.. Why even check exists? shouldn't is_directory only be true if it does exist?

dedmen:
> I intentionally didn't auto create, because I didn't want people to be spammed with logs if they don't need them 

Thats still my blocker. I don't want people to be spammed with logs they don't need, creating logs directory manually was the flag to enable intercept logging.

Maybe we can only create the dir if -debug start parameter is set?
https://github.com/dedmen/ArmaScriptProfiler/blob/master/src/scriptProfiler.cpp#L818-L845

Still, even if they set it once we'll create logs dir and they'll be spammed with logs even if they remove it again

Dahlgren:
Sure, I'll fix it

PR #240, opened 2020-05-25
Build x64 for Linux

Dahlgren:


dedmen:
need to fix up loader scanning for x64, mainly vtable offset

Dahlgren:
For anyone intending to use this, there are still things to fix before Linux x64 is usable

PR #246, opened 2021-04-06
update SQF headers to  A3 2.02 stable

Fusselwurm:


Fusselwurm:
ah wait...you'll want an export from a profiling build, right 😓

dedmen:
You can manually patch in the functions from profiling and the name in the header but thats annoying

src/client/headers/client/sqf_pointers_definitions.hpp
```diff
@@ -262,14 +275,10 @@ unary_function __sqf::unary__deletevehicle__object__ret__nothing;
 unary_function __sqf::unary__deletewaypoint__array__ret__nothing;
 unary_function __sqf::unary__detach__object__ret__nothing;
 unary_function __sqf::unary__detectedmines__side__ret__array;
-unary_function __sqf::unary__diag_captureframe__scalar__ret__nothing;
```
dedmen, line 265: like just copy these over

PR #247, opened 2021-04-28
Update almost everything to v2.05

Leopard20:
Almost every command and event handler up to v2.05 (latest dev/diag build) has been added.
I also fixed several event handlers and some commands

There are only very few (and mostly useless) commands left (see: https://github.com/intercept/intercept/issues/248)

The only exceptions are: hashmaps and diary_records, whose types are not implemented yet.


src/client/intercept/client/sqf/vehicles.cpp
```diff
@@ -918,5 +943,152 @@ namespace intercept {
...
+
+        void enable_vehicle_sensor(const object& vehicle_, sqf_string_const_ref component_, bool enable_) {
+            host::functions.invoke_raw_binary(__sqf::binary__enablevehiclesensor__object__array__ret__nothing, vehicle_, {component_, enable_});
+        }
+
+        std::vector<rv_vehicle_sensor> is_vehicle_sensor_enabled(const object &vehicle_, sqf_string_const_ref component_) {
+            return __helpers::__convert_to_vector<rv_vehicle_sensor>(host::functions.invoke_raw_binary(__sqf::binary__isvehiclesensorenabled__object__string__ret__array, vehicle_, component_));
+        }
+
+        #define RETURN_PYLON_MAGS if (game_return.is_empty()) return sqf_return_string_list(); \
```
Leopard20, line 985: @dedmen plz review this

src/client/headers/client/eventhandlers.hpp
```diff
@@ -331,68 +355,75 @@ namespace intercept::client {
 #define EH_Func_Args_Object_WeaponDisassembled types::object unit, types::object primaryBag, types::object secondarybag
 #define EH_Func_Args_Object_WeaponDeployed types::object unit, bool isDeployed
 #define EH_Func_Args_Object_WeaponRested types::object unit, bool isRested
+#define EH_Func_Args_Object_Disassembled types::object entity, types::object primaryBag, types::object secondaryBag
+#define EH_Func_Args_Object_PathCalculated types::object agent, std::vector<types::vector3> path
```
dedmen, line 359: auto_array would be the native engine one. In most cases i guess native is preferred.
But here we have to copy the values anyway so it doesn't make a difference

src/client/headers/client/eventhandlers.hpp
```diff
@@ -516,37 +551,169 @@ namespace intercept::client {
...
+#define EH_Func_Args_Ctrl_HTMLLink types::control ctrl, types::r_string url
+#define EH_Func_Args_Ctrl_IMEChar types::control ctrl, int char_code
+#define EH_Func_Args_Ctrl_IMEComposition types::control ctrl, int char_code
+#define EH_Func_Args_Ctrl_KeyDown types::control ctrl, int keyCode, bool shift_held, bool ctrl_held, bool alt_held
+#define EH_Func_Args_Ctrl_KeyUp types::control ctrl, int keyCode, bool shift_held, bool ctrl_held, bool alt_held
+#define EH_Func_Args_Ctrl_KillFocus types::control ctrl
+#define EH_Func_Args_Ctrl_LBDblClick types::control ctrl, int selected_index
+    class event_handlers_listbox_info {
+    public:
+        types::r_string text_lb{};
```
dedmen, line 578: put the strings together and the float at end.
strings are x64bit and float is 32bit. Better have the padding at end instead of in the middle

src/client/headers/client/eventhandlers.hpp
dedmen: you used auto_array in EH_Func_Args_Ctrl_LBDrag too. maybe better to stay with engine type

src/client/headers/client/sqf/ai.hpp
```diff
@@ -146,6 +168,7 @@ namespace intercept {
             AUTOTARGET,    // Prevent the unit from assigning a target independently and watching unknown objects / no automatic target selection
             MOVE,          // Disable the AI's movement / do not move
             ANIM,          // Disable ability of AI to change animation.
+            WEAPONAIM,     // Disabled weapon aiming.
```
dedmen, line 171: is this correct enum value really?

dedmen: No its not, but we also don't care I think.

src/client/headers/client/sqf/core.hpp
```diff
@@ -19,6 +19,67 @@ using namespace intercept::types;
 
 namespace intercept {
     namespace sqf {
+        struct rv_mod_info {
+            std::string modname{};
```
dedmen, line 23: sqf string type

src/client/headers/client/sqf/core.hpp
```diff
@@ -19,6 +19,67 @@ using namespace intercept::types;
...
+                  dlcAppId(gv_[1]),
+                  sessionTime(gv_[2]),
+                  totalTime(gv_[3]) {}
+        };
+        struct rv_dlc_asset_info {
+            bool isDlc;                 // - the asset belongs to a DLC
+            bool isOwned;               //
+            bool isInstalled;           //
+            bool isAvailable;           //
+            sqf_return_string appID;    // - actual steam item ID or "0" for none or "-1" for unknown
```
dedmen, line 73: Is the type used in structs really sqf_return_string? the return one?
Isn't that only specifically for returned values?

are there other structs wuth strings? what do they use.

dedmen: https://github.com/intercept/intercept/blob/fd16da9ff8b4b890d51af7c94b86f8804e7b4460/src/client/headers/shared/client_types.hpp#L131

sqf_string.
use sqf_string when storing strings in structs

src/client/headers/client/sqf/debug.hpp
```diff
@@ -38,5 +76,27 @@ namespace intercept {
         void enable_diag_legend(bool value_);
         void halt();  //only in dev version - at least wiki says so
         void throw_exception(game_value exception_);
+        void diag_enable(debug_toggles mode_, bool enable_);
```
dedmen, line 79: I'm not sure if we want to support diag-binary only commands..
I think we don't. People will try to call these while running non-diag, and probably get crashes because these commands don't exist.

src/client/headers/client/sqf/units.hpp
```diff
@@ -17,6 +17,11 @@ using namespace intercept::types;
 
 namespace intercept {
     namespace sqf {
+        struct rv_zeroing {
```
dedmen, line 20: its defined in this file, but not used here?

src/client/headers/client/sqf_pointers_declaration.hpp
```diff
@@ -1,4 +1,4 @@
-//Exported Pointer Definitions For: Arma 3 200.146796 - Profile
+//Exported Pointer Definitions For: Arma 3 205.147535 - Diag
```
dedmen, line 1: Yeah better stay on profile, not diag only commands

src/client/intercept/client/sqf/core.cpp
```diff
@@ -467,6 +467,11 @@ namespace intercept {
                 __sqf::unary__allvariables__control__ret__array, value_));
         }
 
+        sqf_return_string_list all_variables(const display &value_) {
+            return __helpers::__convert_to_vector<sqf_return_string>(host::functions.invoke_raw_unary(
+                __sqf::unary__allvariables__display__ret__array, value_));
+        }
+
         bool __sqfassert(bool test_) {
             return host::functions.invoke_raw_unary(__sqf::unary__isnull__script__ret__bool, test_);
```
dedmen, line 476: lol wtf is this :D

src/client/intercept/client/sqf/debug.cpp
```diff
@@ -82,5 +82,190 @@ namespace intercept {
...
+                case intercept::sqf::debug_draw_modes::Buoyancy:
+                    draw_mode = "Buoyancy"sv;
+                    break;
+                default:
+                    return;
+            }
+            host::functions.invoke_raw_unary(__sqf::unary__diag_drawmode__string__ret__nothing, std::move(draw_mode));
+        }
+
+        void diag_dump_script_assembly(const code& code_) {
```
dedmen, line 189: returns string. But should be removed as all the other diag commands

src/client/intercept/client/sqf/intersects.cpp
```diff
@@ -66,15 +66,16 @@ namespace intercept {
             return __helpers::__line_intersects_surfaces(intersects_value);
         }
 
-        intersect_surfaces_list line_intersects_surfaces(const vector3 &begin_pos_asl_, const vector3 &end_pos_asl_, const object &ignore_obj1_, const object &ignore_obj2_, bool sort_mode_, int max_results_, sqf_string_const_ref lod1_, sqf_string_const_ref lod2_) {
+        intersect_surfaces_list line_intersects_surfaces(const vector3 &begin_pos_asl_, const vector3 &end_pos_asl_, const object &ignore_obj1_, const object &ignore_obj2_, bool sort_mode_, int max_results_, sqf_string_const_ref lod1_, sqf_string_const_ref lod2_, bool unique_objs_) {
```
dedmen, line 69: maybe default =false?

src/client/intercept/client/sqf/vehicles.cpp
dedmen: Seems fine, if you tested it to be working

dedmen (changes requested):
See comments

src/client/intercept/client/sqf/intersects.cpp
Leopard20: It already exists in the declaration in .hpp. isn't that enough?

src/client/intercept/client/sqf/debug.cpp
Leopard20: meh I just added them for completeness

src/client/intercept/client/sqf/core.cpp
Leopard20: did I add that?!

src/client/headers/client/sqf/core.hpp
Leopard20: roger

src/client/headers/client/sqf/core.hpp
Leopard20: sure

src/client/headers/client/sqf/ai.hpp
Leopard20: I use them later in a switch and convert them to string view

src/client/headers/client/sqf/units.hpp
Leopard20: shoot! I forgot to add the declaration! :D
it's used in units.cpp

src/client/headers/client/eventhandlers.hpp
Leopard20: should I do this for all classes/structs that I added?

src/client/intercept/client/sqf/core.cpp
Leopard20: I didn't xD

Leopard20:
I think it should be ok now.

dedmen:
diag commands also make it harder for others to update the pointer list, because that means only people who run diag binary can generate the list

CMakeSettings.json
```diff
@@ -0,0 +1,27 @@
+﻿{
```
dedmen, line 1: This is a local only file, it shouldn't be here

src/client/headers/client/eventhandlers.hpp
dedmen: probably not. Here it makes sense, others it might not

src/client/headers/client/sqf/units.hpp
dedmen: doesn't need to be in header if its not visible to anyone besides units.cpp

src/client/headers/shared/vector.hpp
```diff
@@ -80,12 +81,22 @@ namespace intercept {
             constexpr T magnitude() const noexcept { return std::sqrt(x * x + y * y + z * z); }
             constexpr T magnitude_squared() const noexcept { return x * x + y * y + z * z; }
             constexpr T dot(const vector3_base& v) const noexcept { return (x * v.x + y * v.y + z * v.z); }
+            constexpr T cos(const vector3_base& v) const noexcept { T m = magnitude() * v.magnitude(); if (m == 0.0f) return 0.0f; return dot(v) / m; }
```
dedmen, line 84: Weren't these done in that other commit that was already merged?

src/client/intercept/client/sqf/debug.cpp
dedmen: Well either have them correct, or remove them.
I'm still for removing them as calling them won't work for the majority of users.

src/client/intercept/client/sqf/intersects.cpp
dedmen: ah it is yes

src/client/intercept/client/sqf/misc.cpp
```diff
@@ -5,7 +5,7 @@
 
 namespace intercept {
     namespace sqf {
-
+
```
dedmen, line 8: ??

dedmen (changes requested):

CMakeSettings.json
Leopard20: I pushed it by mistake :(

src/client/intercept/client/sqf/misc.cpp
Leopard20: I'd added something then moved it to another file. I'll remove it

src/client/headers/shared/vector.hpp
Leopard20: wait, when did I add this?

src/client/intercept/client/sqf/debug.cpp
Leopard20: > work for the majority of users.

Well, not everyone uses the profiling build either. I'll remove some completely useless ones tho.

Leopard20:
Let me re-create this branch. It was too messy.

PR #250, opened 2021-05-06
Update commands to v2.05

Leopard20:
Almost all commands up to v2.05 (diag/dev) has been added.
I will update the event handlers in a separate pull request (I want to remove the "randomness" we discussed too).


Leopard20:
@dedmen 
> 
> > Also I removed the diag dump commands because they were indeed useless.
> 
> Which dump commands? they are just as useful as the other diag commands?

dump terrain svg, dump script assembly, etc. I'd added them myself so don't worry.

> 
> > btw not sure why you said it didn't make sense to rearrange the variables in other structs.
> 
> effort and time

They're already done. No more "effort and time" required.

dedmen:
Why do you keep closing and reopening the PR?

Leopard20:
Actually, I'm gonna merge it anyway.

PR #251, opened 2021-05-07
Fixed vector normalize

Leopard20:
1. It didn't handle the case where magnitude could be zero (in Arma it returns [0,0,0] in this case)
2. Why take the abs of a positive number? it was pointless.


Leopard20:
BTW I realized I could just `return *this`. I'll improve that later.

PR #252, opened 2021-05-07
Update and fix event handlers

Leopard20:
All event handlers have been updated.

In addition, the "random number generation" method for UIDs has been changed. The previous method could have collisions and was unsafe. The new one guarantees that you can successively create up to `33'554'432` event handlers of a certain type (object, ctrl, display, mission, MP) in one mission with no collisions.

Of course, the new method is not perfect either. For a perfect solution, you need to know the event handler ID before creating it, which is not possible in Arma at the moment.

### DIFF
--- a/src/client/intercept/client/eventhandlers.cpp
+++ b/src/client/intercept/client/eventhandlers.cpp
@@ -26,27 +26,32 @@ namespace intercept::client {
                 if (found != funcMapMissionEH.end())
                     retVal = callEHHandler(found->second.first, args, found->second.second);
             } break;
-            case eventhandler_type::object: {
+
+			case eventhandler_type::object: {
                 auto found = funcMapObjectEH.find({uid, handle, EHIteration, ehType });
                 if (found != funcMapObjectEH.end())
                     retVal = callEHHandler(found->second.first, args, found->second.second);
             } break;
-            case eventhandler_type::ctrl: {
+
+			case eventhandler_type::ctrl: {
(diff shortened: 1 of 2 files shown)


src/client/intercept/client/eventhandlers.cpp
```diff
@@ -82,14 +82,15 @@ namespace intercept::client {
             case eventhandlers_mission::PreloadStarted: //[[fallthrough]]
             case eventhandlers_mission::PreloadFinished: //[[fallthrough]]
             case eventhandlers_mission::EachFrame: //[[fallthrough]]
+            case eventhandlers_mission::MPEnded: //[[fallthrough]]
             case eventhandlers_mission::Draw3D: {
                 (*func)();
-            }
                 break;
+            }
```
dedmen, line 89: Why move the break around? I prefer it after the } like it was before.
Actually usually on the same line like
`} break;`

Leopard20: I don't know. It looked weird

dedmen:
Besides the moved breaks everywhere (which don't fit code style in other files where we do the same), this looks good..

dedmen:
For pull requests, make a new branch.
When you need to make changes make a new commit on that branch and push it, and done.
Don't do things on master branch

Leopard20:
> 
> 
> Besides the moved breaks everywhere (which don't fit code style in other files where we do the same), this looks good..

I'll put them back.

PR #253, opened 2021-05-27
Add operations between vector2 and vector3

Leopard20:
* Added operations between vector2 and vector3 (z is assumed 0 for vector2).
* Fixed a mistake in `dir_to` member functions

### DIFF
--- a/src/client/headers/shared/vector.hpp
+++ b/src/client/headers/shared/vector.hpp
@@ -102,21 +102,21 @@ namespace intercept {
             constexpr T cos(const vector2_base<T>& v) const noexcept { T m = magnitude() * v.magnitude(); if (m == 0.0f) return 0.0f; return dot(v) / m; }
             constexpr T angle(const vector3_base& v) const noexcept { return std::acos(cos(v)) / pi * 180.0f; }
             constexpr T angle(const vector2_base<T>& v) const noexcept { return std::acos(cos(v)) / pi * 180.0f; }
-            constexpr T distance(const vector3_base& v) const noexcept { vector3_base(*this - v).magnitude(); }
-            constexpr T distance(const vector2_base<T>& v) const noexcept { vector3_base(*this - v).magnitude(); }
-            constexpr T distance_squared(const vector3_base& v) const noexcept { vector3_base(*this - v).magnitude_squared(); }
-            constexpr T distance_squared(const vector2_base<T>& v) const noexcept { vector3_base(*this - v).magnitude_squared(); }
-            constexpr T distance_2d(const vector3_base& v) const noexcept { vector2_base<T>(v - *this).magnitude(); }
-            constexpr T distance_2d(const vector2_base<T>& v) const noexcept { vector2_base<T>(v - *this).magnitude(); }
-            constexpr T distance_2d_squared(const vector3_base& v) const noexcept { vector2_base<T>(v - *this).magnitude_squared(); }
-            constexpr T distance_2d_squared(const vector2_base<T>& v) const noexcept { vector2_base<T>(v - *this).magnitude_squared(); }
+            constexpr T distance(const vector3_base& v) const noexcept { return vector3_base(*this - v).magnitude(); }
+            constexpr T distance(const vector2_base<T>& v) const noexcept { return vector3_base(*this - v).magnitude(); }
+            constexpr T distance_squared(const vector3_base& v) const noexcept { return vector3_base(*this - v).magnitude_squared(); }
+            constexpr T distance_squared(const vector2_base<T>& v) const noexcept { return vector3_base(*this - v).magnitude_squared(); }
+            constexpr T distance_2d(const vector3_base& v) const noexcept { return vector2_base<T>(v - *this).magnitude(); }
+            constexpr T distance_2d(const vector2_base<T>& v) const noexcept { return vector2_base<T>(v - *this).magnitude(); }
+            constexpr T distance_2d_squared(const vector3_base& v) const noexcept { return vector2_base<T>(v - *this).magnitude_squared(); }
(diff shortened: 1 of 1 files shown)


src/client/headers/shared/vector.hpp
```diff
@@ -81,16 +97,26 @@ namespace intercept {
...
             constexpr T angle(const vector3_base& v) const noexcept { return std::acos(cos(v)) / pi * 180.0f; }
-            constexpr T distance(const vector3_base& v) const noexcept { vector3_base dist = (*this - v); dist = dist * dist; return std::sqrt(dist.x + dist.y + dist.z); }
-            constexpr T distance_squared(const vector3_base& v) const noexcept { vector3_base dist = (*this - v); dist = dist * dist; return (dist.x + dist.y + dist.z); }
-            constexpr T distance_2d(const vector3_base& v) const noexcept { vector3_base dist = (*this - v); dist = dist * dist; return std::sqrt(dist.x + dist.y); }
-            constexpr T distance_2d_squared(const vector3_base& v) const noexcept { vector3_base dist = (*this - v); dist = dist * dist; return (dist.x + dist.y); }
+            constexpr T angle(const vector2_base<T>& v) const noexcept { return std::acos(cos(v)) / pi * 180.0f; }
+            constexpr T distance(const vector3_base& v) const noexcept { return vector3_base(*this - v).magnitude(); }
+            constexpr T distance(const vector2_base<T>& v) const noexcept { return vector3_base(*this - v).magnitude(); }
```
dedmen, line 106: Would make more sense to cast vec3 down to vec2, then do 2D delta and 2D magnitude instead of pulling your 0 Z value along the whole way.


Edit:
I think this function is confusing. Distance to a 2D vector should only be in 2D, as you don't know the 2D's Z height.
So taking height difference into account is weird and seems unexpected to me. Which is why I wrote that above as I assumed it would be 2D distance as you only have the 2D data.

src/client/headers/shared/vector.hpp
```diff
@@ -183,15 +218,21 @@ namespace intercept {
...
             constexpr T cos(const vector2_base& v) const noexcept { T m = magnitude() * v.magnitude(); if (m == 0.0f) return 0.0f; return dot(v) / m; }
+            constexpr T cos(const vector3_base<T>& v) const noexcept { T m = magnitude() * v.magnitude(); if (m == 0.0f) return 0.0f; return dot(v) / m; }
             constexpr T angle(const vector2_base& v) const noexcept { return std::acos(cos(v)) / pi * 180.0f; }
-            constexpr T distance(const vector2_base& v) const noexcept { vector2_base dist = (*this - v); dist = dist * dist; return std::sqrt(dist.x + dist.y); }
-            constexpr T distance_squared(const vector2_base& v) const noexcept{ vector2_base dist = (*this - v); dist = dist * dist; return (dist.x + dist.y); }
+            constexpr T angle(const vector3_base<T>& v) const noexcept { return std::acos(cos(v)) / pi * 180.0f; }
+            constexpr T distance(const vector2_base& v) const noexcept { return vector2_base(*this - v).magnitude(); }
+            constexpr T distance(const vector3_base<T>& v) const noexcept { return vector2_base(*this - v).magnitude(); }
```
dedmen, line 227: Okey this is extra confusing.

so 
vector3.distance(vector2) is 3D distance.
vector2.distance(vector3) is 2D distance??

I think its better to just leave these out. The user can always just cast by himself and do whatever he wants to do (explicitly).

src/client/headers/shared/vector.hpp
```diff
@@ -183,15 +218,21 @@ namespace intercept {
...
+            constexpr T distance_squared(const vector2_base& v) const noexcept{ return vector2_base(*this - v).magnitude_squared(); }
+            constexpr T distance_squared(const vector3_base<T>& v) const noexcept { return vector2_base(*this - v).magnitude_squared(); }
```
dedmen, line 229: Again

vector3.distance_squared(vector2) is 3D distance.
vector2.distance_squared(vector3) is 2D distance.

src/client/headers/shared/vector.hpp
```diff
@@ -183,15 +218,21 @@ namespace intercept {
...
             constexpr vector2_base normalize() const noexcept { T m = magnitude(); if (m == 0) return {0.f, 0.f}; return (*this / m); }
             constexpr vector2_base rotate(const T& rad) const noexcept { return vector2_base(x * std::cos(rad) - y * std::sin(rad), y * std::cos(rad) + x * std::sin(rad)); }
             constexpr vector2_base rotate_deg(const T& deg) const noexcept { return rotate(deg * pi / 180.0f); }
             constexpr bool zero_distance() const noexcept { return ((x == 0.0f && y == 0.0f) ? true : false); }
-            constexpr vector2_base dir_to(const vector2_base& v) const noexcept { return normalize(v - *this); }
+            constexpr vector2_base dir_to(const vector2_base& v) const noexcept { return vector2_base(v - *this).normalize(); }
+            constexpr vector2_base dir_to(const vector3_base<T>& v) const noexcept { return vector2_base(v - *this).normalize(); }
```
dedmen, line 235: same here.

vector3.dir_to(vector2) is 3D dir.
vector2.dir_to(vector3) is 2D dir.

src/client/headers/shared/vector.hpp
Leopard20: > as you don't know the 2D's Z height.

>  //Allow operations between vector2 and vector3; in the case of vector2, we always assume z = 0 (which is why operator/ was not added)

I assume it's 0

src/client/headers/shared/vector.hpp
Leopard20: basically my reasoning behind it was that:
vector3 operator (vector2) -> 3D
vector2 operator (vector2) -> 2D

PR #262, opened 2021-12-29
linux_x64: Fix sizings, build, and game-state finding

mcoffin:
This fixes the operability of using intercept on 64-bit linux.

* A few things were different sizes
* the layout of the partial gamestate struct used to find the gamestate pointer was incorrect
* removed unnecessary forcing of compiler options (like `-std`) in places where it doesn't make sense in favor of [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html)
* Use `add_compile_definitions` instead of compiler-implementation-specific define flags
* Fix the linux implementation of `IsBadReadPtr`
* Namings of plugin `.so` filenames did not have the `_x64` suffix in `search.cpp`

The IsBadReadPtr impl works as follows: on Linux, the `write()` implementation for `/dev/random` will *not* segfault when attempting to read a bad pointer. Instead, it will set `errno` to `EACCESS`. We can use this to test the validity of pointers. The previous implementation just did not work at all, and reported that all pointers were valid. I'm not sure that this behavior is necessarily guaranteed by the specification, but reading online showed that others are using it frequently for this purpose as well, so any patch to the kernel that would break this is *highly* unlikely to ever be merged due to breaking userspace usages.

I can confirm that on both 64-bit Windows and 64-bit Linux that this branch is working properly for me at the moment, but I have *not* tested 32-bit compatibility, so if someone would help test that before we lick a stamp and send it, that would be nice :)

I also moved `spdlog` here to `system` to make it easier to use different versions of `spdlog`. This is necessary for running in tandem with `extDB3` during transition periods to avoid segfaults due to differing signatures for spdlog functions. See [`mcoffin-upstream-local`](https://github.com/mcoffin/intercept/tree/mcoffin-upstream-local) for `spdlog` from our currently-utilized `extDB3`. There is also a cmake option (`USE_SYSTEM_SPDLOG`) to use the *system*'s spdlog (`/usr/include`, etc.), but people aren't likely to do this due to some packages not shipping the bundled `fmt` implementation, causing missing includes. Nevertheless, having the option present for using versions of `spdlog` that still have `#include <header.h>` style include directives is a nice feature for those who want to easily drop in their own implementation

Maintaining and rebasing this is obviously not *super* ideal, so if this could actually get upstreamed at some point I think it would be a boon for both me and others running linux dedicated servers.


mcoffin:
While the `intercept-database` side is not ready yet, for those interested, my progress branch is [here](https://github.com/mcoffin/intercept-database/tree/mcoffin). You'll just need to symlink `intercept` to this branch instead of using the included submodule, and fix the hard-coded `-m32 -march=blahblah` stuff in the other submodule-ized dependencies.

src/client/headers/client/sqf/world.hpp
```diff
@@ -233,7 +233,7 @@ namespace intercept {
                 float refractCoef_,
                 float refractSaturation_,
                 bool snow_,
-                bool dropColorStrong_) : rainDropTexture(rainDropTexture_),
+                bool dropColorStrong_) : rainDropTexture(static_cast<std::string_view>(rainDropTexture_)),
```
Leopard20, line 236: Why was this changed? I don't see any reason for casting it to a string_view.
It has to be stored as a string anyway. So using string_view provides no advantage over const string& as far as I see.

src/client/CMakeLists.txt
```diff
@@ -12,7 +12,8 @@ option(USE_STATIC_LINKING "USE_STATIC_LINKING" ON)
 
 
 if((CMAKE_CXX_COMPILER_ID MATCHES "Clang") OR (CMAKE_CXX_COMPILER_ID MATCHES "GNU"))
-    SET(CMAKE_CXX_FLAGS "-std=c++1z -O2 -s -fPIC -fpermissive")
+    # SET(CMAKE_CXX_FLAGS "-std=c++1z -O2 -s -fPIC -fpermissive")
+    add_compile_options("-fPIC" "-fpermissive")
```
Leopard20, line 16: Shouldn't -O2 be kept? Or is it the default compile optimization? I don't use gcc or clang so idk.

src/host/invoker/invoker.cpp
```diff
@@ -65,7 +65,7 @@ namespace intercept {
 
     bool invoker::invoker_end_register(const arguments&, std::string&) {
         //#TODO move pre_start to here
-        _intercept_registerTypes_function.clear(); //#TEST
+        // _intercept_registerTypes_function.clear(); //#TEST
```
dedmen, line 68: why?

src/host/intercept_dll/CMakeLists.txt
```diff
@@ -27,16 +27,14 @@ set_target_properties(${INTERCEPT_EXTENSION_NAME} PROPERTIES FOLDER intercept)
...
 		"$<$<CONFIG:Release>:-O3>"
-		"-s" #remove -s to enable debug symbols. Also maybe add -g or -g3
+		"$<$<CONFIG:Release>:-s>"
 		"-fPIC"
 		"-fpermissive"
 		"-pthread"
 		"-static-libgcc"
 		"-static-libstdc++"
 		"-Wno-ignored-attributes"
-		"-static"
```
dedmen, line 39: why?

src/host/extensions/search.cpp
```diff
@@ -60,8 +61,12 @@ namespace intercept::search {
     std::optional<std::string> plugin_searcher::find_extension(const std::string& name) {
         LOG(INFO, "Searching for Extension: {}", name);
         for (auto folder : active_mod_folder_list) {
+#if defined(__amd64__) || defined(__x86_64__)
+            std::string test_path = folder + "/intercept/" + name + "_x64.so";
+#else
             std::string test_path = folder + "/intercept/" + name + ".so";
-
+#endif
+            std::cerr << "intercept::host::extensions: searching path: " << test_path << std::endl;
```
dedmen, line 69: Debug logging shouldn't be enabled in release builds

.gitignore
```diff
@@ -30,3 +30,6 @@ cmake_install.cmake
 install_manifest.txt
 compile_commands.json
 CTestTestfile.cmake
+
+# Built pbo
+/intercept_core.pbo
```
dedmen, line 35: Pbo's are built inside rv/addons folder (the exclude for that is already above), shouldn't need to do this

src/client/CMakeLists.txt
dedmen: We want to keep -s in release builds. And I don't see why you would remove O2?

src/host/loader/loader.cpp
```diff
@@ -10,13 +12,67 @@
...
+        inline void dump(char prefix) {
+            fprintf(stderr, "\t%c_ptr: %p\n\t%c_count: %d\n\t%c_capacity: %d\n", prefix, ptr, prefix, count, prefix, capacity);
+        }
+    };
+
+    typedef struct gamestate_array<int> gamestate_array_t;
+
+    struct gamestate_partial {
+#if defined(_WIN32) || defined(_WIN64)
+        struct gamestate_array<uintptr_t> types;
```
dedmen, line 43: count/capacity types are definitely in.
But due to engine layout being

struct X1 {
void* data;
int size
}

struct array : X1 {
int capacity
}


it may add padding between them? But they are definitely `int`

With padding there may be random data in the upper 32bits which might fail the equals check

Types is just auto_array

functions/operators/nulars are https://github.com/intercept/intercept/blob/fd16da9ff8b4b890d51af7c94b86f8804e7b4460/src/client/headers/shared/containers.hpp#L1705
Which is very much definitely int, not uintptr_t

src/host/loader/ptraccess.cpp
```diff
@@ -0,0 +1,44 @@
...
+#include <cerrno>
+
+#ifdef __CUSTOM_ISBADREADPTR
+#include <unistd.h>
+#include <fcntl.h>
+#include <optional>
+
+class TmpFd {
+public:
+	TmpFd(const char *path, int mode):
```
dedmen, line 12: This uses tabs.
Intercept uses 4-space indent

dedmen (changes requested):

src/host/loader/loader.cpp
```diff
@@ -10,13 +12,67 @@
...
 namespace intercept {
+    template<typename count_t>
+    struct gamestate_array {
+        void *ptr;
+        count_t count;
+        count_t capacity;
+
+        inline bool is_valid() {
+            if (IsBadReadPtr(ptr, sizeof(void *))) return false;
+            return capacity >= count;
```
dedmen, line 31: At this point, all of them should have equal capacity/count

src/client/headers/client/sqf/world.hpp
mcoffin: This was due to an ambiguity the GCC/Clang don't like in terms of choosing which variant of the constructor to use. Since the `string_view` is constructible from whatever `rainDropTexture_` was, some compilers complain that they can't choose which overloaded version of the constructor to use. If it's better to cast to `const string&` though, then we can do that as well.

src/client/CMakeLists.txt
mcoffin: In this case, `-O2` was being added to *all* builds, including debug builds, as it was being added to the global `CMAKE_CXX_FLAGS`. By default, when using the proper `-DCMAKE_BUILD_TYPE=Release` flag to configure a release build, `cmake` will use `-O3` by default.

Unless the intent was to keep `-O2` for *all* builds, in which case I can just re-add it?

For reference, with `-DCMAKE_BUILD_TYPE=Release`, here's an example of one of the compilation commands that resulted (with the current CMakeLists.txt from this branch).

```
[50/58] /usr/bin/c++ -DDEVEL -DINTERCEPT_HOST_DLL=1 -DNOMINMAX -D_LINUX64=1 -D_STATIC -I/home/mcoffin/projects/intercept/build-release2/src/host/extensions -I/home/mcoffin/projects/intercept/src/host/extensions -I/home/mcoffin/projects/intercept/src/host/common -I/home/mcoffin/projects/intercept/src/host/controller
 -I/home/mcoffin/projects/intercept/src/host/loader -I/home/mcoffin/projects/intercept/src/host/invoker -I/home/mcoffin/projects/intercept/src/host/../client/headers -I/home/mcoffin/projects/intercept/src/host/intercept_dll -I/home/mcoffin/projects/intercept/src/host/common/system -I/home/mcoffin/projects/intercept
/build-release2/src/host/common -I/home/mcoffin/projects/intercept/build-release2/src/host/controller -I/home/mcoffin/projects/intercept/build-release2/src/host/loader -I/home/mcoffin/projects/intercept/build-release2/src/host/invoker -I/home/mcoffin/projects/intercept/build-release2/src/host/intercept_dll -Wno-att
ributes -O3 -DNDEBUG -fPIC -std=c++17 -MD -MT src/host/extensions/CMakeFiles/extensions_static.dir/extensions.cpp.o -MF src/host/extensions/CMakeFiles/extensions_static.dir/extensions.cpp.o.d -o src/host/extensions/CMakeFiles/extensions_static.dir/extensions.cpp.o -c /home/mcoffin/projects/intercept/src/host/extens
ions/extensions.cpp
```

So, it looks like `cmake` by default will add `-DNDEBUG -O3` for the release build type.

Either way, we shouldn't be setting release-specific flags via `CMAKE_CXX_FLAGS`, and instead `CMAKE_CXX_FLAGS_RELEASE`, or conditionals as is done with the target-specific flags below.

src/host/invoker/invoker.cpp
mcoffin: Forgot to un-comment after testing. Good catch will update.

src/host/loader/loader.cpp
mcoffin:
```cpp
            struct size_check {
                uintptr_t p1;//typearray
                uintptr_t typeCount;
                uintptr_t typeCapacity;
                uintptr_t p2;//functions
                int f_tableCount{ 0 };
                int f_count{ 0 };
                uintptr_t p3;//operators
                int o_tableCount{ 0 };
                int o_count{ 0 };
                uintptr_t p4;//nulars
                int n_tableCount{ 0 };
                int n_count{ 0 };
            };
```
snipped for reference ^

src/host/loader/loader.cpp
mcoffin: Ah, so these are all guaranteed to be compact arrays, gotcha. Will update. :+1:

src/host/loader/ptraccess.cpp
mcoffin: :+1:

src/client/headers/client/sqf/world.hpp
mcoffin: Actually, it appears from my commit log that this was actually MSVC that was complaining about the ambiguity in which overloaded version of the constructor to use.

src/host/loader/loader.cpp
mcoffin: as long as you think that the new logic is the "more correct" one, we can for sure just remove the templated type and move it all over, I was just minimizing risk of breaking someone else's build in case I had a misunderstanding

src/host/loader/loader.cpp
mcoffin: This was *not* the case for me locally (default branch `x64` linux), so we likely still have some typing issues with maybe garbage data in higher bits somewhere as mentioned above. I'll start looking in to it.

src/host/intercept_dll/CMakeLists.txt
dedmen: If something seems to make no difference, I would prefer to leave it as it was

src/host/loader/loader.cpp
dedmen: Engine calls Compact on these, after commands were added, Which should ensure that capacity==count.

John-H-Smith:
What is blocking right now?
Feature highly appreciated!

dedmen:
> What is blocking right now?
The open comments

src/client/headers/client/sqf/world.hpp
dedmen: You are also breaking the sqf_string passing here.
Instead of being able to pass the string directly, you are forcing a conversion to string view and a copy. Thats not nice.
This really sohuldn't be needed

src/host/loader/loader.cpp
dedmen: If this was wrong, and your uintptr_t were correct, then 
https://github.com/intercept/intercept/blob/6a5a4a914c9a82f2e5e2bc2876185fccafc4a638/src/client/headers/shared/containers.hpp#L1703
would also be wrong

dedmen: devenv_2023-07-07_19-48-05
Yeah your code was wrong, unless you want to say that A3 only has 127 functions/operators

dedmen: Actually I didn't realize how messed up this was, functions, operators, nulars. Aren't arrays at all

PR #265, opened 2022-06-24
Update to v2.09.149590

Leopard20:
I've also added a new "fake" command: `get_tide()`, which is just `getTerrainInfo#4` (sea level)


src/client/headers/client/sqf/world.hpp
```diff
@@ -423,5 +423,24 @@ namespace intercept {
         std::vector<object> nearest_mines(const vector3 &pos_, sqf_string_list_const_ref types_, float radius_, bool sort_ = true, bool mode_2d_ = false);
         std::vector<object> nearest_mines(const vector2 &pos_, sqf_string_list_const_ref types_, float radius_, bool sort_ = true, bool mode_2d_ = false);
 
+        struct rv_terrain_info {
+            float terrain_grid_width;
+            int terrain_grid_size;
+            float sea_level;
```
dedmen, line 429: seaLevel should be last, to match both constructor initialization order and also the ingame array thing

dedmen (approved):

PR #268, opened 2022-07-19
loader: Account for multiple memory regions in maps for linux x64

mcoffin:
This change is required to support the new compiler that began being
used with the 2.08 v18+ versions of the performance branch on linux x64

### DIFF
--- a/src/host/loader/loader.cpp
+++ b/src/host/loader/loader.cpp
@@ -38,7 +38,7 @@ namespace intercept {
 #if _WIN32 || _WIN64
         explicit MemorySection(const MODULEINFO& modInfo) noexcept :
             start(reinterpret_cast<uintptr_t>(modInfo.lpBaseOfDll)),
-            end(reinterpret_cast<uintptr_t>(modInfo.SizeOfImage) + reinterpret_cast<uintptr_t>(modInfo.lpBaseOfDll)) {}
+            end(static_cast<uintptr_t>(modInfo.SizeOfImage) + reinterpret_cast<uintptr_t>(modInfo.lpBaseOfDll)) {}
 #endif // _WIN32 || _WIN64
         MemorySection(uintptr_t _start, uintptr_t _end) noexcept : start(_start), end(_end) {}
         uintptr_t start;
--- a/src/host/loader/loader.cpp
+++ b/src/host/loader/loader.cpp
@@ -2,6 +2,11 @@
 #include "controller.hpp"
 #include <thread>
 #include <future>
+#include <iostream>
+#include <cstdio>
+#include <cstring>
+#include <optional>
+#include <regex>
 #ifdef __linux__
 #include <dlfcn.h>
 #include <link.h>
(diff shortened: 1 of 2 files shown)


mcoffin:
Note that at least *I personally* still run the changes in #262 to achieve the remainder of linux/x64 compatibility (game_state pointer finding changes), so I was unable to re-test this after I rebased the commit off of `master`, but the rebase was relatively simple.

The original commit can be seen at mcoffin/intercept@1efc5dc52f72d62e8d39a7f54f617fc15e5de254

@dedmen If you want to give the windows logic branch a once-over, it should be pretty much 100% the same, just using one memory region as before, that would be nice as I don't currently have a working windows setup. The logic should be 100% the same, just using the `MemorySections` class with only one section (effectively the same as before).

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,127 @@
...
+                #ifdef LOADER_DEBUG
+                    std::cerr << "intercept::loader: map line: " << line << std::endl;
+                    std::cerr << '\t' << std::hex << startParsed << "-" << std::hex << endParsed << std::endl;
+                #endif
+
+                // Add section to array
+                sections.push_back(MemorySection(reinterpret_cast<void*>(startParsed), reinterpret_cast<void*>(endParsed)));
+            }
+        }
+        explicit MemorySections(uintptr_t _start, uintptr_t _size): sections({MemorySection(reinterpret_cast<void*>(_start), reinterpret_cast<void*>(_start + _size))}) {}
```
mcoffin, line 121: FYI - constructor used by windows logic to create a section list with only one entry, with the global start and end from `GetModuleInfo`

src/host/loader/loader.cpp
```diff
@@ -78,61 +197,33 @@ namespace intercept {
...
-            return 0;
+        #ifdef __linux__
+            MemorySections memorySections("/proc/self/maps");
+        #else
+            MODULEINFO modInfo = { nullptr };
+            HMODULE hModule = GetModuleHandleA(nullptr);
+            GetModuleInformation(GetCurrentProcess(), hModule, &modInfo, sizeof(MODULEINFO));
+            const uintptr_t baseAddress = reinterpret_cast<uintptr_t>(modInfo.lpBaseOfDll);
+            const uintptr_t moduleSize = static_cast<uintptr_t>(modInfo.SizeOfImage);
+            MemorySections memorySections(baseAddress, moduleSize);
```
mcoffin, line 216: Since it's using just one memory region, this *should* result in the exact same behavior as before, but it could use a quick test.

src/host/loader/loader.cpp
```diff
@@ -196,17 +290,35 @@ namespace intercept {
...
-        #else
-            uintptr_t vtableStart = stringOffset - (0x09D20C70 - 0x09D20BE8);
-            return vtableStart;
-            //return (findInMemory(reinterpret_cast<char*>(&vtableStart), 4));
-        #endif
-
+            #ifndef __linux__
+                return (findInMemory(reinterpret_cast<char*>(&stringOffset), sizeof(uintptr_t)) - sizeof(uintptr_t));
+            #elif _LINUX64
+                bool oldCompiler = *reinterpret_cast<uintptr_t*>(stringOffset - 8) == stringOffset;
```
mcoffin, line 296: This section of logic is

```
Co-authored-by: dedmen <dedmen@dedmen.de>
```

mcoffin:
Tested this with [`futurecept`](https://github.com/olympus-entertainment/futurecept), and all appears well, indicating that downstream things are still working. Thanks for the help, and let me know if you need anything else from me!

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,127 @@
...
+                    }
+                }
+                if (found) {
+                    return base + i;
+                }
+            }
+            return std::nullopt;
+        }
+        std::optional<uintptr_t> findInMemoryPattern(const char* pattern, const char* mask, uintptr_t offset = 0) const {
+            const uintptr_t base = reinterpret_cast<const uintptr_t>(start);
```
dedmen, line 62:
```suggestion
        MemorySection(void* _start, void* _end) noexcept : start( reinterpret_cast<uintptr_t>(_start)), end( reinterpret_cast<uintptr_t>(_end)) {}
        uintptr_t start;
        uintptr_t end;
        inline size_t size() const {
            return end - start;
        }
        std::optional<uintptr_t> findInMemory(const char* pattern, size_t patternLength) const {
            const uintptr_t base = start;
            const auto sz = size();
            for (uintptr_t i = 0; i < sz - patternLength; i++) {
                bool found = true;
                for (uintptr_t j = 0; j < patternLength; j++) {
                    found &= pattern[j] == *reinterpret_cast<char*>(base + i + j);
                    if (!found) {
                        break;
                    }
                }
                if (found) {
                    return base + i;
                }
            }
            return std::nullopt;
        }
        std::optional<uintptr_t> findInMemoryPattern(const char* pattern, const char* mask, uintptr_t offset = 0) const {
            const uintptr_t base = start;
```
May aswell store them right away as uintptr_t, if every usage of these is trying to convert them

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,127 @@
...
+            }
+            return std::nullopt;
+        }
+    };
+
+    class MemorySections {
+    protected:
+        std::vector<MemorySection> sections;
+    public:
+        explicit MemorySections(const char* path = "/proc/self/maps") {
```
dedmen, line 85: #ifdef __linux__ for this linux only code

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,127 @@
...
+            for (auto section = sections.cbegin(); section != sections.cend(); section++) {
+                auto found = section->findInMemory(pattern, patternLength);
+                if (found.has_value()) {
+                    return found.value();
+                }
+            }
+            return static_cast<uintptr_t>(0);
+        }
+        uintptr_t findInMemory(const std::string_view& pattern) const {
+            return findInMemory(&pattern.front(), pattern.length());
```
dedmen, line 132:
```suggestion
            return findInMemory(pattern.data(), pattern.length());
```

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,127 @@
...
+        }
+        uintptr_t findInMemoryPattern(const char* pattern, const char* mask, uintptr_t offset = 0) const {
+            for (auto section = sections.cbegin(); section != sections.cend(); section++) {
+                auto found = section->findInMemoryPattern(pattern, mask, offset);
+                if (found.has_value()) {
+                    return found.value();
+                }
+            }
+            return static_cast<uintptr_t>(0x0u);
```
dedmen, line 141: Why the static cast if just return 0 would do it too?
And would be just as obvious to understand and easier to read

mcoffin: I just left it from the previous iteration of `findInMemory(Pattern)?`, but I agree. Will remove

src/host/loader/loader.cpp
dedmen: This could also be moved into default constructor, and then #ifdef windows'ed there

src/host/loader/loader.cpp
dedmen: oh yeah if I wrote that I wrote stupid stuff :D

src/host/loader/loader.cpp
mcoffin: done :)

src/host/loader/loader.cpp
mcoffin: done :)

src/host/loader/loader.cpp
mcoffin: I mean the *real* answer would be to make an `std::initializer_list` constructor for `MemorySections`, and do

```cpp
MemorySections memorySections {MemorySection(baseAddress, moduleSize)};
```

But, my c++ isn't the *best* on the planet, and `std::initializer_list` ambiguity in constructors tends to make my head roll.

It probably doesn't need to be `#ifdef windows`'d , because it *could* still be a useful constructor on linux for creating a instance that only references one memory section, but that's just my opinion, don't think that one matters too much

mcoffin: Let me know if this blocks this PR, because I can change it, but personally I think that should probably just be left without the `#ifdef` in case we ever need it for searching only single sections at some point in the future.

mcoffin: Just in case, addressed this with c5d27d334d2c4c607db5413c41c9f0df8638ea36. Feel free to include that (or not) when merging. I agree it does look nicer to get rid of the extra `#ifdef` later on.

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,144 @@
...
+            end(reinterpret_cast<uintptr_t>(modInfo.SizeOfImage) + reinterpret_cast<uintptr_t>(modInfo.lpBaseOfDll)) {}
+#endif // _WIN32 || _WIN64
+        MemorySection(uintptr_t _start, uintptr_t _end) noexcept : start(_start), end(_end) {}
+        uintptr_t start;
+        uintptr_t end;
+        inline size_t size() const {
+            return end - start;
+        }
+        std::optional<uintptr_t> findInMemory(const char* pattern, size_t patternLength) const {
+            const uintptr_t base = reinterpret_cast<const uintptr_t>(start);
```
dedmen, line 50:
```suggestion
            const uintptr_t base = start;
```

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,144 @@
...
+                    }
+                }
+                if (found) {
+                    return base + i;
+                }
+            }
+            return std::nullopt;
+        }
+        std::optional<uintptr_t> findInMemoryPattern(const char* pattern, const char* mask, uintptr_t offset = 0) const {
+            const uintptr_t base = reinterpret_cast<const uintptr_t>(start);
```
dedmen, line 67:
```suggestion
            const uintptr_t base = start;
```

dedmen (approved):

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,144 @@
...
+                // Parse out begin/end memory locations
+                auto startParsed = std::strtoul(match[1].str().c_str(), nullptr, 16);
+                auto endParsed = std::strtoul(match[2].str().c_str(), nullptr, 16);
+                #ifdef LOADER_DEBUG
+                    std::cerr << "intercept::loader: map line: " << line << std::endl;
+                    std::cerr << '\t' << std::hex << startParsed << "-" << std::hex << endParsed << std::endl;
+                #endif
+
+                // Add section to array
+                sections.push_back(MemorySection(reinterpret_cast<uintptr_t>(startParsed), reinterpret_cast<uintptr_t>(endParsed)));
```
dedmen, line 134:
```suggestion
                sections.push_back(MemorySection(static_cast<uintptr_t>(startParsed), static_cast<uintptr_t>(endParsed)));
```

src/host/loader/loader.cpp
```diff
@@ -16,13 +21,144 @@
...
     loader::~loader() {
-
     }
 
+    class MemorySection {
+    public:
+#if _WIN32 || _WIN64
+        explicit MemorySection(const MODULEINFO& modInfo) noexcept :
+            start(reinterpret_cast<uintptr_t>(modInfo.lpBaseOfDll)),
+            end(reinterpret_cast<uintptr_t>(modInfo.SizeOfImage) + reinterpret_cast<uintptr_t>(modInfo.lpBaseOfDll)) {}
```
dedmen, line 41:
```suggestion
            end(static_cast<uintptr_t>(modInfo.SizeOfImage) + reinterpret_cast<uintptr_t>(modInfo.lpBaseOfDll)) {}
```

PR #275, opened 2023-02-23
Added support for hashmap type

Leopard20:
To-do: 
- [x] fix parameter name styles to conform to intercept's style (e.g. `var` -> `var_`)
- [x] add an ~~unordered_map~~ (created rv_hashmap for now; use with `__helper::__convert_to_hashmap`) conversion so we can return from commands
- [x] add missing sqf commands that returned hashmaps
- [x] ~~maybe rename `key` and `value` to `first` and `second`?~~ no need

Sample test (RSQF):
SQF:
```sqf
testFn "ppBloom"
```
```cpp
game_value test_fn(game_value_parameter _id) {
    auto vidOpts = get_video_options();
    for (auto& opt : vidOpts) {
        system_chat(str({ opt.key, opt.value }));
    }
    auto v = vidOpts.find(_id);
    if (v)
        return v->value;
    else
        return "not found";
}
```

### DIFF
--- a/src/client/headers/client/sqf/common_helpers.hpp
+++ b/src/client/headers/client/sqf/common_helpers.hpp
@@ -1,4 +1,4 @@
-﻿/*!
+/*!
 @file
 @author Verox (verox.averre@gmail.com)
 @author Nou (korewananda@gmail.com)
@@ -70,6 +70,10 @@ namespace intercept {
                 if (input_.size() == 0) return {};
                 auto& arr = input_.to_array();
                 return std::vector<T>(arr.begin(), arr.end());
+            }
+
+            inline rv_hashmap __convert_to_hashmap(const game_value& input_) {
+                return std::move(input_.get_as<game_data_hashmap>()->data);
             }
 
 
--- a/src/client/headers/shared/client_types.hpp
+++ b/src/client/headers/shared/client_types.hpp
@@ -490,13 +490,9 @@ namespace intercept {
             obj_type_temporary = 4,
             obj_type_vehicle = 8,
             obj_type_temp_veh = 16,
-            obj_type_land_decal = 32,
-            obj_type_all = 63,
-            obj_type_all_inc_null = -1
+            obj_type_land_decal = 32
         };
-#ifdef DEFINE_ENUM_FLAG_OPERATORS
-        DEFINE_ENUM_FLAG_OPERATORS(object_types)
-#endif
+
         enum class object_simulation_kind {
             Statics = 0,
             Slow = 0,
--- a/src/client/headers/shared/client_types.hpp
+++ b/src/client/headers/shared/client_types.hpp
@@ -523,7 +523,7 @@ namespace std {
     
 #define DEFINE_HASH_FUNCTION_FOR_CLASS(type) template <> struct hash<intercept::types::type> { \
         size_t operator()(const intercept::types::type& x) const { \
-            return x.hash(); \
+            return x.get_hash(); \
         } \
     }; 
 
--- a/src/client/headers/shared/client_types.hpp
+++ b/src/client/headers/shared/client_types.hpp
@@ -490,9 +490,13 @@ namespace intercept {
             obj_type_temporary = 4,
             obj_type_vehicle = 8,
             obj_type_temp_veh = 16,
-            obj_type_land_decal = 32
+            obj_type_land_decal = 32,
+            obj_type_all = 63,
+            obj_type_all_inc_null = -1
         };
-
+#ifdef DEFINE_ENUM_FLAG_OPERATORS
+        DEFINE_ENUM_FLAG_OPERATORS(object_types)
+#endif
         enum class object_simulation_kind {
             Statics = 0,
             Slow = 0,
--- a/src/client/headers/shared/containers.hpp
+++ b/src/client/headers/shared/containers.hpp
@@ -1654,6 +1654,7 @@ namespace intercept::types {
         return hashValue;
     }
 
+
     struct map_string_to_class_trait {
         /*
         static unsigned int hash_key(const char* key) noexcept {
@@ -1663,6 +1664,7 @@ namespace intercept::types {
             return strcmp(k1, k2) == 0;
         }
         */
+
         static unsigned int hash_key(std::string_view key) noexcept {
             return rv_map_hash_string_case_sensitive(key);
         }
@@ -1801,7 +1803,8 @@ namespace intercept::types {
             }
         }
 
-        const Type& get(std::string_view key_) const {
+        template <class T>
+        const Type& get(T&& key_) const {
             if (!_table || !_count) return _null_entry;
             const int hashed_key = hash_key(key_);
             for (size_t i = 0; i < _table[hashed_key].count(); i++) {
@@ -1812,19 +1815,22 @@ namespace intercept::types {
             return _null_entry;
         }
 
-        Container* get_table_for_key(std::string_view key_) {
+        template <class T>
+        Container* get_table_for_key(T&& key_) {
             if (!_table || !_count) return nullptr;
             const int hashed_key = hash_key(key_);
             return &_table[hashed_key];
         }
 
-        const Container* get_table_for_key(std::string_view key_) const {
+        template <class T>
+        const Container* get_table_for_key(T&& key_) const {
             if (!_table || !_count) return nullptr;
             const int hashed_key = hash_key(key_);
             return &_table[hashed_key];
         }
 
-        Type& get(std::string_view key_) {
+        template <class T>
+        Type& get(T&& key_) {
             if (!_table || !_count) return _null_entry;
             const int hashed_key = hash_key(key_);
             for (size_t i = 0; i < _table[hashed_key].count(); i++) {
@@ -1837,7 +1843,8 @@ namespace intercept::types {
 
         static bool is_null(const Type& value_) { return &value_ == &_null_entry; }
 
-        bool has_key(std::string_view key_) const {
+        template <class T>
+        bool has_key(T&& key_) const {
             return !is_null(get(key_));
         }
 
@@ -1910,7 +1917,8 @@ namespace intercept::types {
             return x;
         }
 
-        bool remove(std::string_view key) {
+        template <class T>
+        bool remove(T&& key) {
             if (_count <= 0) return false;
 
             int hashedKey = hash_key(key);
@@ -1926,12 +1934,14 @@ namespace intercept::types {
         }
 
         //Is empty?
-        bool empty() {
+        bool empty() const {
             return (!_table || !_count);
         }
 
     protected:
-        int hash_key(std::string_view key_) const {
+
+        template <class T>
+        int hash_key(T&& key_) const {
             return Traits::hash_key(key_) % _tableCount;
         }
     };
--- a/src/client/headers/shared/containers.hpp
+++ b/src/client/headers/shared/containers.hpp
@@ -1664,7 +1664,7 @@ namespace intercept::types {
             return strcmp(k1, k2) == 0;
         }
         */
-
+        using key_type = std::string_view;
         static unsigned int hash_key(std::string_view key) noexcept {
             return rv_map_hash_string_case_sensitive(key);
         }
@@ -1700,6 +1700,8 @@ namespace intercept::types {
                               k2.begin(), [](unsigned char l, unsigned char r) { return l == r || tolower(l) == tolower(r); });
         }
     };
+
+
 
     template <class Type, class Container, class Traits = map_string_to_class_trait>
     class map_string_to_class {
@@ -1803,8 +1805,7 @@ namespace intercept::types {
             }
         }
 
-        template <class T>
-        const Type& get(T&& key_) const {
+        const Type& get(typename Traits::key_type key_) const {
             if (!_table || !_count) return _null_entry;
             const int hashed_key = hash_key(key_);
             for (size_t i = 0; i < _table[hashed_key].count(); i++) {
@@ -1815,22 +1816,19 @@ namespace intercept::types {
             return _null_entry;
         }
 
-        template <class T>
-        Container* get_table_for_key(T&& key_) {
+        Container* get_table_for_key(typename Traits::key_type key_) {
             if (!_table || !_count) return nullptr;
             const int hashed_key = hash_key(key_);
             return &_table[hashed_key];
         }
 
-        template <class T>
-        const Container* get_table_for_key(T&& key_) const {
+        const Container* get_table_for_key(typename Traits::key_type key_) const {
             if (!_table || !_count) return nullptr;
             const int hashed_key = hash_key(key_);
             return &_table[hashed_key];
         }
 
-        template <class T>
-        Type& get(T&& key_) {
+        Type& get(typename Traits::key_type key_) {
             if (!_table || !_count) return _null_entry;
             const int hashed_key = hash_key(key_);
             for (size_t i = 0; i < _table[hashed_key].count(); i++) {
@@ -1843,8 +1841,7 @@ namespace intercept::types {
 
         static bool is_null(const Type& value_) { return &value_ == &_null_entry; }
 
-        template <class T>
-        bool has_key(T&& key_) const {
+        bool has_key(typename Traits::key_type key_) const {
             return !is_null(get(key_));
         }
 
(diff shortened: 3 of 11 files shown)


src/client/headers/shared/types.hpp
```diff
@@ -554,13 +561,13 @@ namespace intercept {
                 copy(copy_);
             }
 
-#pragma optimize("ts", on)
+//#pragma optimize("ts", on)
```
killerswin2, line 564: #pragma optimize( "ts", on ) is equivalent to /O.

src/client/headers/shared/client_types.hpp
```diff
@@ -490,9 +490,13 @@ namespace intercept {
             obj_type_temporary = 4,
             obj_type_vehicle = 8,
             obj_type_temp_veh = 16,
-            obj_type_land_decal = 32
```
dedmen, line 493: Unrelated change to hashmap, separate PR

src/client/headers/shared/containers.hpp
```diff
@@ -1801,7 +1803,8 @@ namespace intercept::types {
             }
         }
 
-        const Type& get(std::string_view key_) const {
+        template <class T>
+        const Type& get(T&& key_) const {
```
dedmen, line 1807: Should get key type from traits.
Though only argument I can think of is compilation speedup because fewer template methods, and better syntax autocomplete info thing when writing code.
Because here it doesn't know what type you could pass in.

src/client/headers/shared/containers.hpp
```diff
@@ -1801,7 +1803,8 @@ namespace intercept::types {
...
             if (!_table || !_count) return _null_entry;
             const int hashed_key = hash_key(key_);
```
dedmen, line 1809: if you want to do perfect forwarding, then you're missing the std::forward here

src/client/headers/shared/types.hpp
```diff
@@ -69,6 +70,7 @@ namespace intercept {
             NetObject,
             SUBGROUP,
             TEAM_MEMBER,
+            HASHMAP,
```
dedmen, line 73: Shouldn't this be at the bottom? instead of adding a new thing randomly in the middle?

src/client/headers/shared/types.hpp
dedmen: Unrelated change, shouldn't be in this PR

src/client/headers/shared/types.hpp
```diff
@@ -685,11 +694,11 @@ namespace intercept {
             uintptr_t get_vtable() const noexcept {
                 return *reinterpret_cast<const uintptr_t*>(this);
             }
-#pragma optimize("ts", on)
+//#pragma optimize("ts", on)
```
dedmen, line 697: unrelated change

src/client/headers/shared/types.hpp
```diff
@@ -903,6 +912,73 @@ namespace intercept {
...
+
+            inline game_hashmap_type::iterator begin() { return data.begin(); }
+            inline game_hashmap_type::iterator end() { return data.end(); }
+
+            pair* find(const game_value& key);
+            const pair* find(const game_value& key) const;
+            bool contains(const game_value& key) const;
+
+            inline int size() const { return data.count(); }
+            inline bool empty() const { return data.empty(); }
```
dedmen, line 973: Instead of reimplementing all of this.
just make to_hashmap return the actual hashmap instead of the game data
So users have direct access to the actual hashmap

src/client/headers/shared/types.hpp
```diff
@@ -1589,7 +1665,7 @@ namespace std {
     template <>
     struct hash<intercept::types::game_value> {
         size_t operator()(const intercept::types::game_value& x) const {
-            return x.hash();
+            return x.data->get_hash();
```
dedmen, line 1668: data might be nullptr
Better to implement a hash/get_hash method in game_value and handle it there

dedmen (changes requested):

src/client/headers/shared/types.hpp
Leopard20: uh I forgot these! It was because I'm using the Struct Layout extension and it uses Clang for compilation, and Clang just freaks about about that pragma saying it should be defined in file-scope...

src/client/headers/shared/types.hpp
Leopard20: Yeah that's a better idea...

src/client/headers/shared/types.hpp
Leopard20: right didn't notice that. will do

src/client/headers/shared/types.hpp
Leopard20: I didn't mean to remove these. But I'll wrap them in a `#ifdef __clang__` in another PR

src/client/headers/client/sqf/common_helpers.hpp
```diff
@@ -1,4 +1,4 @@
-﻿/*!
+/*!
```
dedmen, line 1: UTF8 bom header?

src/client/headers/shared/containers.hpp
```diff
@@ -1900,23 +1935,19 @@ namespace intercept::types {
...
-        bool remove(std::string_view key) {
+        bool remove(typename Traits::key_type key) {
             if (_count <= 0) return false;
 
             int hashedKey = hash_key(key);
-            for (size_t i = 0; i < _table[hashedKey].size(); i++) {
+            for (size_t i = 0; i < _table[hashedKey].size(); ++i) {
                 Type& item = _table[hashedKey][i];
-                if (Traits::compare_keys(item.get_map_key(), key) == 0) {
+                if (Traits::compare_keys(item.get_map_key(), key)) {
```
dedmen, line 1950: did it always return bool? as the old code was ==0?

src/client/headers/client/sqf/common_helpers.hpp
Leopard20: visual studio added that 😓

src/client/headers/shared/containers.hpp
```diff
@@ -1926,19 +1957,100 @@ namespace intercept::types {
...
-        bool empty() {
+        bool empty() const {
             return (!_table || !_count);
         }
 
+        void reserve(int newTable) {
+            if (newTable <= _tableCount)
+                return;
+
+            rebuild(newTable);
```
dedmen, line 1968:
```suggestion
        void reserve(int newTableCount) {
            if (newTableCount <= _tableCount)
                return;

            rebuild(newTableCount);
```

src/client/headers/shared/containers.hpp
Leopard20: yeah. you must've confused it with `k1.compare(k2) == 0` in here:
```cpp
static bool compare_keys(std::string_view k1, std::string_view k2) noexcept {
            if (k1.length() != k2.length()) return false;
            return k1.compare(k2) == 0;
        }
        ```

File: src/client/headers/shared/containers.hpp (hunk 11 above)
Comment (dedmen) [reply src/client/headers/shared/containers.hpp:1968]: Also, reserve is usually for number of elements, not number of tables.
But we don't actually implement expansion factor at all so...... eh..? 

Either keep this, or make it number of entries and calculate like.. one table per 10 entries or smth..
But we'd want it log scale and.. ugh

File: src/client/headers/shared/containers.hpp [hunk 12]
```diff
@@ -1926,19 +1957,100 @@ namespace intercept::types {
...
+        const typename Type::value_type& at(typename Type::key_type key_) const {
+            auto& temp = this->get(key_);
+            if (this->is_null(temp)) {
+                throw std::out_of_range("Invalid key");
+            }
+            return temp.value;
+        }
+
+        template<class...Args>
+        Type& emplace(Args&&...args_) {
```
dedmen, line 2028:
```suggestion
        template<class... Args>
        Type& emplace(Args&&... args_) {
```

src/client/headers/shared/containers.hpp
Leopard20: everywhere else it is also just `compare_keys` as bool:
```cpp
if (Traits::compare_keys(item.get_map_key(), key_))
```

src/client/headers/shared/containers.hpp
```diff
@@ -1926,19 +1957,100 @@ namespace intercept::types {
...
+            return this->insert(Type(std::forward<Args>(args_)...));
+        }
+
+        Type* find(typename Type::key_type key_) {
+            auto& temp = this->get(key_);
+            if (this->is_null(temp))
```
dedmen, line 2034: is null can be static.
also shouldn't need `this->`?

dedmen: yeah why the this-> everywhere?

src/client/headers/shared/types.hpp
dedmen: So its my fault then :D

src/client/headers/shared/types.hpp
```diff
@@ -898,11 +909,51 @@ namespace intercept {
             static void* operator new(std::size_t sz_);
             static void operator delete(void* ptr_, std::size_t sz_);
             bool val;
-            size_t hash() const { return __internal::pairhash(type_def, val); }
+            size_t hash() const { return this->get_hash(); }
```
dedmen, line 912:
```suggestion
            [[deprecated("Use get_hash instead")]] size_t hash() const { return this->get_hash(); }
```

src/client/headers/shared/types.hpp
```diff
@@ -878,7 +889,7 @@ namespace intercept {
             static void operator delete(void* ptr_, std::size_t sz_);
             float number;
             size_t hash() const {
```
dedmen, line 891:
```suggestion
            [[deprecated("Use get_hash instead")]] size_t hash() const {
```

src/client/headers/shared/types.hpp
```diff
@@ -920,7 +971,7 @@ namespace intercept {
             ~game_data_array();
             auto_array<game_value> data;
             auto length() const { return data.count(); }
-            size_t hash() const { return __internal::pairhash(type_def, data.hash()); }
+            size_t hash() const { return this->get_hash(); }
```
dedmen, line 974:
```suggestion
            [[deprecated("Use get_hash instead")]] size_t hash() const { return this->get_hash(); }
```

src/client/headers/shared/types.hpp
```diff
@@ -939,7 +990,7 @@ namespace intercept {
             game_data_string& operator=(game_data_string&& move_) noexcept;
             ~game_data_string();
             r_string raw_string;
-            size_t hash() const { return __internal::pairhash(type_def, raw_string); }
+            size_t hash() const { return this->get_hash(); }
```
dedmen, line 993:
```suggestion
            [[deprecated("Use get_hash instead")]] size_t hash() const { return this->get_hash(); }
```

src/client/headers/shared/types.hpp
```diff
@@ -954,7 +1005,7 @@ namespace intercept {
                 *reinterpret_cast<uintptr_t*>(this) = type_def;
                 *reinterpret_cast<uintptr_t*>(static_cast<I_debug_value*>(this)) = data_type_def;
             }
-            size_t hash() const { return __internal::pairhash(type_def, group); }
+            size_t hash() const { return this->get_hash(); }
```
dedmen, line 1008:
```suggestion
            [[deprecated("Use get_hash instead")]] size_t hash() const { return this->get_hash(); }
```

src/client/headers/shared/types.hpp
```diff
@@ -966,7 +1017,7 @@ namespace intercept {
                 *reinterpret_cast<uintptr_t*>(this) = type_def;
                 *reinterpret_cast<uintptr_t*>(static_cast<I_debug_value*>(this)) = data_type_def;
             }
-            size_t hash() const { return __internal::pairhash(type_def, config); }
+            size_t hash() const { return this->get_hash(); }
```
dedmen, line 1020:
```suggestion
            [[deprecated("Use get_hash instead")]] size_t hash() const { return this->get_hash(); }
```

src/client/headers/shared/types.hpp
```diff
@@ -979,7 +1030,7 @@ namespace intercept {
                 *reinterpret_cast<uintptr_t*>(this) = type_def;
                 *reinterpret_cast<uintptr_t*>(static_cast<I_debug_value*>(this)) = data_type_def;
             }
-            size_t hash() const { return __internal::pairhash(type_def, control); }
+            size_t hash() const { return this->get_hash(); }
```
dedmen, line 1033:
```suggestion
            [[deprecated("Use get_hash instead")]] size_t hash() const { return this->get_hash(); }
```

src/client/headers/client/sqf/common_helpers.hpp
Leopard20: but I guess that's harmless right?

src/client/headers/shared/types.hpp
```diff
@@ -991,7 +1042,7 @@ namespace intercept {
                 *reinterpret_cast<uintptr_t*>(this) = type_def;
                 *reinterpret_cast<uintptr_t*>(static_cast<I_debug_value*>(this)) = data_type_def;
             }
-            size_t hash() const { return __internal::pairhash(type_def, display); }
+            size_t hash() const { return this->get_hash(); }
```
dedmen, line 1045:
```suggestion
            [[deprecated("Use get_hash instead")]] size_t hash() const { return this->get_hash(); }
```

src/client/headers/shared/types.hpp
```diff
@@ -1003,7 +1054,7 @@ namespace intercept {
                 *reinterpret_cast<uintptr_t*>(this) = type_def;
                 *reinterpret_cast<uintptr_t*>(static_cast<I_debug_value*>(this)) = data_type_def;
             }
-            size_t hash() const { return __internal::pairhash(type_def, location); }
+            size_t hash() const { return this->get_hash(); }
```
dedmen, line 1057:
```suggestion
           [[deprecated("Use get_hash instead")]]  size_t hash() const { return this->get_hash(); }
```


Actually now after making all these suggestions.
You can just plainly remove all these, and add

`[[deprecated("Use get_hash instead")]]  size_t hash() const { return this->get_hash(); }`
directly into game_data instead

src/client/intercept/shared/types.cpp
```diff
@@ -821,46 +887,19 @@ namespace intercept {
...
-#endif
-                case game_data_type::TEAM_MEMBER: return reinterpret_cast<game_data_team_member*>(data.get())->hash();
-                case game_data_type::TASK: return reinterpret_cast<game_data*>(data.get())->to_string().hash(); //"Task %s (id %d)" or "No Task"
-                case game_data_type::DIARY_RECORD: return reinterpret_cast<game_data*>(data.get())->to_string().hash(); //"No diary record" or... The text of that record? Text might be long and make this hash heavy
-                case game_data_type::LOCATION: return reinterpret_cast<game_data_location*>(data.get())->hash();
-                case game_data_type::end: return 0;
-            }
+        uint64_t game_value::get_hash() const {
+            if (!data)
+                return 0;
```
dedmen, line 897: empty fnv-1a hash is not zero, but... doesn't matter

src/client/intercept/shared/types.cpp
```diff
@@ -821,46 +887,19 @@ namespace intercept {
...
+            return data.get()->get_hash();
+        }
 
-            return types::__internal::pairhash<uintptr_t,uintptr_t>(data->get_vtable(),reinterpret_cast<uintptr_t>(data.get()));
+        size_t game_value::hash() const {
```
dedmen, line 901: If its just a copy for back compat, could make it deprecated. but doesn't matter, probably wouldn't bother

src/client/headers/shared/types.hpp
Leopard20: Yeah idk what I was thinking 😅

src/client/headers/shared/containers.hpp
Leopard20: visual studio was complaining can't find `get`, etc. when I was compiling. doing that fixed it 🤷‍♂️

src/client/headers/shared/containers.hpp
dedmen: It was compare returning number before, but I recently changed it in RV.
Guess I had the same idea back then with intercept.
but...
if it was ==0 before, then that was basically a not so the old logic was inveretd and wrong? :D 
oof, sad

src/client/headers/client/sqf/common_helpers.hpp
dedmen: In many projects, the most changed line is the first, because people kepe back and forth'ing the bom mark :D 
Doesn't matter-ish

Also I can't see if it was wrong before and you fixed it, or if it was right before and you broke it :D

src/client/headers/shared/containers.hpp
Leopard20: yup you were removing non-matching keys :D

src/client/headers/shared/containers.hpp
Leopard20: expansion factor seems to be 16 I guess?

src/client/headers/client/sqf/common_helpers.hpp
Leopard20: Looks like it was wrong before and visual studio fixed it by itself 😅

PR #278, opened 2023-03-15
division operators

killerswin2:
was missing some vector3 operators, namely division with vector 2.
Follows the same idea as multiplication of vec3 by vec2. ie z is 0


Leopard20:
I didn't add them on purpose, as stated a few lines above:

> // Allow operations between vector2 and vector3; in the case of vector2, we always assume z = 0 (which is why operator/ was not added)

killerswin2:
ah okay.

PR #279, opened 2023-03-31
Fix and improve auto_array

Leopard20:
* Fixes the `clear` bug
* Makes `reserve` reserve exactly instead of using grow
* `resize` doesn't deallocate anymore.

### DIFF
--- a/src/client/headers/shared/containers.hpp
+++ b/src/client/headers/shared/containers.hpp
@@ -70,12 +70,12 @@ namespace intercept::types {
 
         static void destroy_deallocate(Type* _Ptr) {
             destroy(_Ptr);
-            return __internal::rv_allocator_deallocate_generic(_Ptr);
+            deallocate(_Ptr);
         }
 
         static void destroy_deallocate(Type* _Ptr, size_t size_) {
             destroy(_Ptr, size_);
-            return __internal::rv_allocator_deallocate_generic(_Ptr);
+            deallocate(_Ptr);
         }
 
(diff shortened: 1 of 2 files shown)


src/client/headers/shared/containers.hpp
```diff
@@ -1232,35 +1232,24 @@ namespace intercept::types {
...
-                return;
-            }
-            newData = Allocator::create_uninitialized_array(size_);
-            //memset(newData, 0, size * sizeof(Type));
+            Type* newData = Allocator::allocate(size_);
             if (base::_data) {
+                if (base::_n) {
 #ifdef __GNUC__
-                memmove(newData, base::_data, base::_n * sizeof(Type));
+                    memmove(newData, base::_data, base::_n * sizeof(Type));
```
dedmen, line 1247: uff this is kinda wrong, not all types are simply movable like that :X 
But its always been this bad so is fine to stay bad.

src/client/headers/shared/containers.hpp
```diff
@@ -1315,28 +1304,22 @@ namespace intercept::types {
 
         auto_array& operator=(const auto_array& copy_) {
             if (copy_._n) {
-                clear();
+                clear(true);
```
dedmen, line 1307: It might be benficial to not clear memory. Then you don't need to reallocate on insert.
But thats rare, and with this you'd have a exactly correct size array which is also nice

Leopard20:
I closed this one. Just see the other one 😄

PR #280, opened 2023-04-01
Fix and improve auto_array

Leopard20:
* Fixes the `clear` bug
* Fixes allowing `erase` past the end of the vector
* Makes `reserve` reserve exactly instead of using `grow`
* `resize` doesn't deallocate anymore.
* `grow` uses a different algorithm that scales with capacity (doubles the capacity as long as it's below a `maxGrowth` number)

~~Also I'm now using `Allocator::allocate` instead of `Allocator::reallocate`. If reallocate was using a pool allocator I guess I shouldn't have done that...~~

### DIFF
--- a/src/client/headers/shared/containers.hpp
+++ b/src/client/headers/shared/containers.hpp
@@ -1347,7 +1347,7 @@ namespace intercept::types {
         */
         template <class... _Valty>
         iterator emplace(iterator where_, _Valty&&... val_) {
-            if (where_ < base::begin() || where_ >= base::end()) throw std::runtime_error("Invalid Iterator");  //WTF?!
+            if (where_ < base::begin() || where_ > base::end()) throw std::runtime_error("Invalid Iterator");  //WTF?!
             const size_t insertOffset = where_ - base::begin();
             auto previousEnd = base::_n;
             if (_maxItems < base::_n + 1) {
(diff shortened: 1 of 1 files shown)


src/client/headers/shared/containers.hpp
```diff
@@ -70,12 +70,12 @@ namespace intercept::types {
...
         static void destroy_deallocate(Type* _Ptr) {
             destroy(_Ptr);
-            return __internal::rv_allocator_deallocate_generic(_Ptr);
+            deallocate(_Ptr);
         }
 
         static void destroy_deallocate(Type* _Ptr, size_t size_) {
             destroy(_Ptr, size_);
-            return __internal::rv_allocator_deallocate_generic(_Ptr);
+            deallocate(_Ptr);
```
dedmen, line 78:
```suggestion
            deallocate(_Ptr, size_);
```

src/client/headers/shared/containers.hpp
```diff
@@ -1232,44 +1232,35 @@ namespace intercept::types {
...
-                return;
-            }
-            newData = Allocator::create_uninitialized_array(size_);
-            //memset(newData, 0, size * sizeof(Type));
+            Type* newData = size_ > 0 ? Allocator::allocate(size_) : nullptr;
             if (base::_data) {
+                if (base::_n) {
 #ifdef __GNUC__
-                memmove(newData, base::_data, base::_n * sizeof(Type));
+                    memcpy(newData, base::_data, base::_n * sizeof(Type));
```
dedmen, line 1247: uff this is kinda wrong, not all types are simply movable like that :X
But its always been this bad so is fine to stay bad.

dedmen (approved):

src/client/headers/shared/containers.hpp
Leopard20: well it didn't do anything with the size so...
but I'll put it there anyway...

src/client/headers/shared/containers.hpp
Leopard20: This is on reallocation though (expanding/shrinking capcaity). Dunno what's wrong with it 🤔 
Do you mean in case something stores a pointer to that var? But still that's the other thing's fault...
I don't really see anything wrong here 🤷‍♂️

Leopard20: But the one thing that can be wrong is changing `Allocator::reallocate` to `Allocator::allocate`.

What did reallocate do differently? Shouldn't I have changed it?

PR #287, opened 2023-07-05
Create callstack_item_data class, for helping place items on the call stack

killerswin2:
creates callstack_item_data class. Moves VM_Context to the end of the file as the containter classes need the full type information else they static assert as failures.


dedmen:
Maybe we should block new/delete on it. To indicate that you shouldn't create a instance of it.
Because it veeery likely won't work right with all the missing stuff

killerswin2:
https://github.com/killerswin2/hash_plugin/blob/master/src/hash.cpp#L983-L1011
https://github.com/killerswin2/hash_plugin/blob/master/src/hash.cpp#L1020-L1048
https://github.com/killerswin2/hash_plugin/blob/master/src/hash.cpp#L1084-L1178
https://github.com/killerswin2/hash_plugin/blob/master/src/hash.cpp#L1249-L1346
I was able to get it to work in a hashmap plugin I was working on. It is possible does seem like a hackish way to do it, but items and variables were getting created and return on the call stack

src/client/headers/shared/types.hpp
```diff
@@ -1354,6 +1258,179 @@ namespace intercept {
...
+                }
+            };
+
+            auto add_callstack_item(ref<callstack_item> newItem)
+            {
+                return callstack.emplace_back(newItem);
+            }
+
+            void throw_script_exception(game_value value)
+            {
```
Leopard20, line 1375:
```suggestion
            void throw_script_exception(game_value value) {
```

src/client/headers/shared/types.hpp
```diff
@@ -1354,6 +1258,179 @@ namespace intercept {
...
+                    file[0] = 0;
+                    line = 0;
+                }
+                virtual IDebugScope* getParent() {
+                    return _parent;
+                }
+            };
+
+            auto add_callstack_item(ref<callstack_item> newItem)
+            {
```
Leopard20, line 1370: The opening braces must be inline with the function declaration:
```suggestion
            auto add_callstack_item(ref<callstack_item> newItem) {
```

src/client/headers/shared/types.hpp
```diff
@@ -1354,6 +1258,179 @@ namespace intercept {
...
+            }
+
+            void throw_script_exception(game_value value)
+            {
+                exception_state = true;
+                exception_value = std::move(value);
+            }
+
+            bool is_scheduled() const
+            {
```
Leopard20, line 1381:
```suggestion
            bool is_scheduled() const {
```

src/client/headers/shared/types.hpp
```diff
@@ -1354,6 +1258,179 @@ namespace intercept {
...
+                return scheduled;
+            }
+
+            bool is_serialization_enabled() const
+            {
```
Leopard20, line 1386:
```suggestion
            bool is_serialization_enabled() const {
```

src/client/headers/shared/types.hpp
```diff
@@ -1354,6 +1258,179 @@ namespace intercept {
...
+                return serialenabled;
+            }
+
+            void disable_serialization()
+            {
```
Leopard20, line 1391:
```suggestion
            void disable_serialization() {
```

src/client/headers/shared/types.hpp
```diff
@@ -1354,6 +1258,179 @@ namespace intercept {
...
+                serialenabled = false;
+            }
+
+            const sourcedocpos& get_current_position()
+            {
```
Leopard20, line 1396:
```suggestion
            const sourcedocpos& get_current_position() {
```

Leopard20 (changes requested):

PR #288, opened 2023-07-13
loader: Fix compilation on Linux after refactor fallout

mcoffin:
intercept/intercept@46b293080f5989b7fb850139c45c979b0c5579c7 still had a few minor things broken for linux 64-bit compatibility. I cleaned up the syntactical stuff, and it *does* compile and work (even with custom types) for at least the **old** versions of performance/profiling from before the command type registration changes.

I'll try it again with one of the latest versions of the `performance` branch later to confirm that types are in fact working for me as well as soon as I can.

Even though the loader preprocessor spam is still pretty gnarly, it'll be *super* nice to not be working off of a constantly-rebasing fork. I'm nervous about how the types changes will go since I had all but given up on using newer versions of performance branch, but if they're coming to stable then I'm going to have to just spend the time and get ahead of it, so thanks for getting some of the previous stuff in!

- [ ] Tested Linux compatibility (including types) with **old** binaries
- [ ] Tested Linux compatibility (including types) with **new** binaries
- [ ] Tested Windows compatibility (including types) with **old** binaries
- [ ] Tested Windows compatibility (including types) with **new** binaries


src/client/headers/shared/containers.hpp
```diff
@@ -1,4 +1,5 @@
 #pragma once
+#include <cstdint>
```
mcoffin, line 2: A recent (last maybe 4 months?) update to GCC started getting mad about this not having been included from elsewhere. I didn't have the time to track it down, so I just slapped 'er right in. `containers.hpp` makes hefty-enough use of those types though, so doesn't seem *that* evil-ly lazy.

src/host/loader/CommandTypesDynamic.hpp
```diff
@@ -74,6 +74,7 @@ namespace intercept {
                     reinterpret_cast<TBase*>(this)->_name2 = newOP;
             }
 
+#ifndef __linux__
```
mcoffin, line 77: These all just remove the (unused on linux) getters since the underlying fields don't exist

src/host/loader/loader.cpp
```diff
@@ -171,6 +171,7 @@ namespace intercept {
         return future_allocatorVtablePtr;
     }
 
+#ifndef __linux__
```
mcoffin, line 174: Here, I just gated the entire existence of `startPoolAllocatorSearch` upon not being on linux, since those servers don't have pool allocators (that I know of).

src/host/loader/loader.cpp
```diff
@@ -256,8 +250,10 @@ namespace intercept {
 
         //#TODO these patternfinds can be replaced by taking the alloc function out of any Types createFunction. and the dealloc function is right next to it asm wise
 
+#ifndef __linux__
```
mcoffin, line 253: This previously didn't compile on linux due to missing futures. Gated the entire thing on `!defined(__linux__)`

mcoffin:
This still needs a few tweaks to not bust 32bit windows builds, so consider it a WIP for now.

src/client/headers/shared/client_types.hpp
```diff
@@ -525,12 +525,14 @@ namespace std {
         reference_wrapper(string& _Val) noexcept
             : reference_wrapper<std::string>(_Val) {};
     };
-    
+
 #define DEFINE_HASH_FUNCTION_FOR_CLASS(type) template <> struct hash<intercept::types::type> { \
         size_t operator()(const intercept::types::type& x) const { \
-            return x.get_hash(); \
+            auto ret = x.get_hash(); \
+            assert(ret < SIZE_MAX); \
```
mcoffin, line 532: Wouldn't this assertion actually fail on 32-bit builds?

If so, then this actually *is* an invalid conversion... but we just want to ignore the data loss?

PR #291, opened 2023-12-21
Loader fixes and improvements

Leopard20:
- [x] Use a better and safer alternative to `IsBadReadPtr`
- [x] Fix loader for vc143
- [x] Fix hashmap poolAlloc


killerswin2:
Linux good?

Leopard20:
> Linux good?

Fixed just now

killerswin2:
lol answered my own question

Leopard20:
Looks like Linux has been failing due to another issue too.

Avokadomos:
This fixed the issues I had with HashMaps on Win64.

src/host/loader/ptraccess.h
```diff
@@ -2,6 +2,46 @@
 #pragma once
 #if defined(_WIN32) || defined(_WIN64)
 #include <windows.h>
+#ifdef __cplusplus
+// From: https://stackoverflow.com/questions/496034/most-efficient-replacement-for-isbadreadptr
+// Check memory address access
+static const DWORD dwForbiddenArea = PAGE_GUARD | PAGE_NOACCESS;
```
dedmen, line 8: Static const should maybe be upper case. Don't know if intercept has other examples of that?

dedmen (approved):

src/host/loader/ptraccess.h
Leopard20: Dunno. I literally copy pasted the code from that stack overflow link 😅 (and added the static)
I should probably replace it with constexpr tho?

PR #292, opened 2024-01-13
Fix loader dealloc on vc143

Leopard20:


src/host/loader/loader.cpp
```diff
@@ -183,7 +183,7 @@ namespace intercept {
...
+                result = memorySections.findInMemoryPattern("\x48\x85\xD2\x0F\x84\x19\x02\x00\x00\x53\x48\x83\xEC\x20\x48\x63\x41\x58\x48\x8B\xD9\x48\xFF\xC8\x48\x89\x7C\x24\x40\x48\x23\xC2\x48\x8B\xFA\x48\x2B\xF8\x83\x3F\x00\x74\x28\x48\x8D\x0D\x96\xAF\x82\x00\xE8\xB9\xE8\xFF\xFF\x44\x8B\x07\x48\x8D\x0D\xAF\xAF\x82\x00\x48\x8B\xD7\x48\x8B\x7C\x24\x40\x48\x83\xC4\x20\x5B\xE9\x6D\xEF\xFF\xFF\x48\x8B\x47\x18\x48\x89\x02\x48\x8B\x4F\x18\x48\x89\x57\x18\x48\x89\x7A\x08\x8B\x57\x10\xFF\xCA\x48\x89\x6C\x24\x30\x33\xED\x89\x57\x10\x48\x85\xC9\x75\x72\x48\x8B\x4F\x28\x4C\x8D\x47\x20\x49\x8B\x00\x48\x89\x01\x49\x8B\x08\x49\x8B\x40\x08\x48\x89\x41\x08\x49\x8B\xC8\x49\x89\x68\x08\x49\x89\x28\xFF\x4B\x40\x85\xD2\x75\x26\x48\x8B\x43\x20\x48\x8D\x53", "xxxxx????xxxxxxxxxxxxxxxxxxx?xxxxxxxxxxxxxxxxx????x????xxxxxx????xxxxxxx?xxxxxx????xxxxxxxxxxxxxxxxxxxxxxxxxxxx?xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx");
```
dedmen, line 186: This isn't correct.

Pattern \x00\x00\x00\x00 changed to \x19\x02\x00\x00
But the mask is still ????, meaning that change has no effect.

Also the pattern is now shorter, but the masks length wasn't changed.

Leopard20: > the mask is still ????, meaning that change has no effect.

So?! 😅 
I just copied it directly from memory

> Also the pattern is now shorter, but the masks length wasn't changed.

It's the same length

PR #295, opened 2024-04-12
Update to v2.17.151686; Fix loader to work with diag exe

Leopard20:
* Added all commands up to v2.17.151686
* Fix loader to work on diag exe


dedmen:
But why does a loader fix, have so many changes to old command wrappers?

dedmen:
I'd prefer to have one PR per thing, not merging multiple together, that makes it harder to review

src/client/intercept/client/sqf/ai.cpp
```diff
@@ -610,9 +610,10 @@ namespace intercept {
             host::functions.invoke_raw_binary(__sqf::binary__landat__object__object_scalar__ret__nothing, value0_, value1_);
         }
 
-        void land_at(const object &value0_, const object &helipad_) {
-            host::functions.invoke_raw_binary(__sqf::binary__landat__object__object_scalar__ret__nothing, value0_, helipad_);
```
dedmen, line 614: Merging this now is annoying. Because this is not available in 2.16.
And intercepts intended usage is to run on stable.
So after this is merged, people on 2.16 stable, will get a null here and cannot use the landAt object syntax anymore.

I doubt anyone is using it, and they could revert to a older revision. But ugh.

PR #298, opened 2024-05-18
Update to v2.17.151825

Leopard20:


src/client/headers/client/sqf/world.hpp
```diff
@@ -448,6 +448,28 @@ namespace intercept {
...
+                sqf_string pp_effect_type;   // one of Normal, NightVision, Thermal, Colors, Mirror, ChromAber or FilmGrain;
+                float TIIndex;               // see setCamUseTI for possible values;
+                explicit r2t(const game_value &gv_) : r2t_name{gv_[0]},
+                                                      r2t_effect_name{gv_[1]},
+                                                      pp_effect_type{gv_[2]},
+                                                      TIIndex{gv_[3]} {}
+            };
+            object camera;              // the camera itself (see camCreate);
+            std::vector<r2t> r2t_infos;  // [r2tName, r2tEffectName, ppEffectType, TIindex]
+            sqf_string effect_name;     // (Optional, only available if isPrimary is true) see cameraEffect for possible values;
```
dedmen, line 466: We coooould make these std::optional's

src/client/headers/client/sqf/world.hpp
```diff
@@ -448,6 +448,28 @@ namespace intercept {
...
+            float view_mode;            // (Optional, only available if isPrimary is true) 0 = normal, 1 = NVG, 2 = TI, 3 = NVG + TI;
```
dedmen, line 467: Might aswell make this enum/bitset?
Not really needed.
In actuality its an int though, same for TIIndex, its an int at the end

dedmen (approved):

src/client/headers/client/sqf/world.hpp
Leopard20: I just copy pasted stuff from the wiki (I used my conversion script that I shared in the Wiki section) 😅 
But yeah enum makes more sense.

